PR #29, opened 2018-01-08
Issue 25

### LINKED ISSUE #25: Pick a recommended encoding
Making people choose between HTML and JSON (as well as XML unless #24 is accepted) decreases interoperability if there is no guidance on which one must be supported. Some may choose HTML, some JSON, some XML. 

It forces them to do additional thinking and research, or just go with their gut - but not matter what there is additional decision making. 

It matters less what the default is, and indeed the second one can be strongly encouraged.

My gut often says HTML, but I've been convinced that it should be JSON. The primary consumer of WFS API's is machines, and it should be a clear machine readable format. HTML should be _strongly_ encouraged, and indeed there likely could be open source libraries that make it really easy to turn the JSON responses in to useful HTML - even doing it on the fly in javascript.

cportele:
* add recommendation on GeoJSON in the Core
* add discussion about encodings and the recommendations in Clause 6
(Overview)
* reference discussion from clauses 2, 7 and 8

resolves #25


cportele:
Link to the text in Clause 6:

https://cdn.rawgit.com/opengeospatial/WFS_FES/issue-25/docs/17-069.html#overview

cportele:
Note: The edits in the PR are based on the edits in PR #21.

cportele:
@pvretano Should we initiate a SWG vote on this as there have been no comments on this PR?

pvretano:
eVote initiated.

PR #34, opened 2018-01-29
support bounding boxes that span the antimeridian

### LINKED ISSUE #31: Bounding Box Terminology for Geographic CRS
7.8.4.18 currently refers to:

> The bounding box is provided as four numbers:
> 1. Lower corner, coordinate axis 1 (minimum longitude)
> 2. Lower corner, coordinate axis 2 (minimum latitude)
> 3. Upper corner, coordinate axis 1 (maximum longitude)
> 4. Upper corner, coordinate axis 2 (maximum latitude)

And the OpenAPI yml also has
```
bbox:
  description: minimum longitude, minimum latitude, maximum longitude, maximum latitude
```

For geographic/polar/cylindrical/etc coordinate systems it's important to use (left, bottom, right, top); or (west, south, east, north) when referring to bounding boxes, since it's clearly possible to have useful bounding boxes that span the discontinuity. 

eg. describing the area of the New Zealand Exclusive Economic Zone; longitudes 160.6°E - 170°W (spanning the antimeridian) & latitudes 55.95°S - 25.89°S

This should be represented as  `{160.6, -55.95, -170, -25.89}`

But using "minimum" and "maximum" terminology leads to lots of software either:

- asserting that x0 < x1 and failing because -170 < 160.6
- attempting to correct "bad" boxes by normalising them (ie. to `{-170, -55.95, 160.6, -25.89}`) so it now describes the the rest of the world **except** the area we want

Alternatively, the spec could support longitudes <-180° and >+180° (currently not permitted), but that moves the problem to the server, since in order to filter the dataset properly the server needs to re-normalise or otherwise adapt them. Or leave it otherwise undefined like most of the other OGC specs.

_[OGC 06-121r3](https://portal.opengeospatial.org/files/?artifact_id=20040
) D.13 Minimum bounding boxes_ discusses this and sits firmly on the fence ("[the min/max approach] can always be used by proper selection of the CRS"), but describes the above WSEN approach as:

> b) For a circular coordinate, specify that the LowerCorner shall define the box edge furthest toward decreasing values, and the UpperCorner shall define the box edge furthest toward larger values. For longitude, the LowerCorner longitude would define the West-most box edge, and the UpperCorner longitude would define the East-most box edge. (The LowerCorner would no longer always use the minimum value, and the UpperCorner would no longer always use the maximum value. The value at the LowerCorner can be greater than at the UpperCorner when this bounding box crosses the value discontinuity.)

cportele:
resolves #31

Note that OpenAPI includes and links still go to the master branch,
i.e. will only be updated once the changes have been merged.

### DIFF
--- a/core/openapi/parameters/bbox.yaml
+++ b/core/openapi/parameters/bbox.yaml
@@ -2,11 +2,17 @@ name: bbox
 in: query
 description: >-
   Only features that have a geometry that intersects the bounding box are selected. The bounding box is provided as four numbers:
-   
-  * Lower corner, coordinate axis 1 (minimum longitude)
-  * Lower corner, coordinate axis 2 (minimum latitude)
-  * Upper corner, coordinate axis 1 (maximum longitude)
-  * Upper corner, coordinate axis 2 (maximum latitude)
+
+  * Left box edge, coordinate axis 1
+  * Bottom box edge, coordinate axis 2
+  * Right box edge, coordinate axis 1
+  * Top box edge, coordinate axis 2
+
+  For WGS84 longitude/latitude this is in most cases the sequence of
+  minimum longitude, minimum latitude, maximum longitude and maximum latitude.
+  However, in cases where the box spans the antimeridian the first value
+  (left = west-most box edge) is larger than the third value (right = east-most
+  box edge).
 required: false
 schema:
   type: array
--- a/core/openapi/parameters/bbox.yaml
+++ b/core/openapi/parameters/bbox.yaml
@@ -3,16 +3,15 @@ in: query
 description: >-
   Only features that have a geometry that intersects the bounding box are selected. The bounding box is provided as four numbers:
 
-  * Left box edge, coordinate axis 1
-  * Bottom box edge, coordinate axis 2
-  * Right box edge, coordinate axis 1
-  * Top box edge, coordinate axis 2
+  * Lower left corner, coordinate axis 1
+  * Lower left corner, coordinate axis 2
+  * Upper right corner, coordinate axis 1
+  * Upper right corner, coordinate axis 2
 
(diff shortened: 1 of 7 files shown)


cportele:
https://rawgit.com/opengeospatial/WFS_FES/issue-31/docs/17-069.html#_parameter_bbox

core/openapi/parameters/bbox.yaml
```diff
@@ -2,11 +2,17 @@ name: bbox
...
-   
-  * Lower corner, coordinate axis 1 (minimum longitude)
-  * Lower corner, coordinate axis 2 (minimum latitude)
-  * Upper corner, coordinate axis 1 (maximum longitude)
-  * Upper corner, coordinate axis 2 (maximum latitude)
+
+  * Left box edge, coordinate axis 1
+  * Bottom box edge, coordinate axis 2
+  * Right box edge, coordinate axis 1
+  * Top box edge, coordinate axis 2
```
rcoup: I _presume_ the coordinate axis order is CRS-dependent? So Axis 1 could either be an Easting _or_ Northing?

In that case, maybe "Lower left corner, coordinate axis 1" & "Lower left corner, coordinate axis 2" is more accurate? (cf Upper right)

Or "SouthWest" & "NorthEast"?

cportele: Yes, I was struggling how to best write it considering that 

* the axis order of the CRS applies and
* WGS84 longitude/latitude is the only CRS that we discuss in the core.

It would be nice, if we can find a wording that is unambiguous, but not (too) confusing for developers that are not familiar with coordinate reference systems. I agree, "lower left corner"/"upper right corner" is better.

cportele:
Agreement on 2018-02-01: Merge immediately

Issue #23, opened 2017-12-13
Paging improvements


cholmes:
So I believe paging as it stands right now is a bit problematic. Hopefully a developer who has actually worked with paging on restful api's with elastic search backends sounds in and can explain more.

But in Planet's data API we don't have 'startIndex' or anything like that, since it's a lot more complexity to enable clients to request arbitrary locations in an index. Indeed if the index changes, with data added or deleted then that can throw off the results.

Planet solves this by requiring users to create a saved search, or there's a shortcut 'quicksearch' to get results right away. But both create an index on the server side that's set in time, that the user can page through. DigitalGlobe actually doesn't enable paging on their Catalog results, and the underlying vector data store seems to require a similar creation of a paging ID - https://gbdxdocs.digitalglobe.com/v1/docs/vs-retrieve-page-of-vector-items

I don't think we need to specify that level of paging / search in the core of the spec (though possibly an extension / best practice). But I think we should remove 'startIndex', in favor of just having results supply their 'next' link, which can be generated by the service. Planet's 'next' links look like https://api.planet.com/data/v1/searches/f8abab5007a14b31b5ccfb8a3d3f02d1/results?_page=eyJxdWVyeV9wYXJhbXMiOiB7fSwgInNvcnRfcHJldiI6IGZhbHNlLCAicGFnZV9zaXplIjogMjUwLCAic29ydF9ieSI6ICJwdWJ

(with a string that is actually 3 times that long). 

In the [STAC spec](http://github.com/radiantearth/stac-spec) we just have an optional nextPageToken, and then the 'links' section uses that token.

Services wouldn't be required to use an obscure string for paging through results. Though we should make a recommendation on if services are expected to return consistent results, or if it is ok that they are returning like less than the 'count'. See like https://developers.facebook.com/blog/post/478/ for an api that doesn't guarantee the number of results.

Additional discussion on paging best practices at https://stackoverflow.com/questions/13872273/api-pagination-best-practices - we should perhaps recommend some default ordering of results that paging can be driven off of. 

The other thing that seems a bit arbitrary is the 10000 maximum limit on the 'count' parameter. And I couldn't figure out in the spec if that is something that implementations can change, or is like hard coded in. At Planet we enable dynamic setting of the page size, but the limit is 250 results per page. And the operations team resists any increases to that, as it introduces more complexity to support.

pvretano:
[...]

Issue #22, opened 2017-12-13
Proposal: Remove 'resultType' from core and put in an extension


cholmes:
(happy to create a PR for the spec changes of this and other issues I'll put in, but I've delayed on getting this feedback out, so figured I'd post now instead of continuing to wait).

Though being able to ask for 'hits' instead of actual results likely seems pretty trivial to implement, it is an additional overhead, and can be more of a pain with a backend like elasticsearch than a standard rdbms. Especially when you get to hundreds of millions of features. At Planet in our data API we have a dedicated 'stats' endpoint that returns the equivalent of 'hits' (see [full api spec](https://api.planet.com/spec) and search for stats). Ours also can return time bucketed data, which drives nice looking graphs. One could see an advanced 'stats' endpoint  being used for [crossfilter](https://square.github.io/crossfilter/) style interactive filtering.

So a good microservices architecture I believe would have 'hits' be its own endpoint that takes the same 'query' as the 'results' endpoint, so it could be backed by a cluster dedicated and optimized for aggregations. Hits as an extension would allow services to include it if it's easy to implement, but not require it for minimal implementations.

For the [SpatioTemporal Asset Catalog spec](https://github.com/radiantearth/stac-spec) we started with a draft WFS 3.0 swagger spec, but removed the result type stuff, see the [swagger spec](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/spec.yaml)

cportele:
I think you are right. Probably we should see how we support resultType=hits and other related capabilities like #13 and #16 consistently in an extension. 

Having a capability to determine the total number of hits is often useful, so we should tackle this soon, if we move it out.

pvretano:
Yeah, I think I agree with this too.  We could probably this about doing something with HEAD to get this kinds of metadata from a server.

cholmes:
Yes, fully agree that the capability to determine the total number of hits is extremely useful - it was one of my favorite additions. 

Just want to be sure it's not required, and that it could easily be implemented in a microservice style architecture. I think there's potential for standardization of even more types of aggregations - like bucketing the stats to drive charts, or even spatial aggregations - return counts by regular geographies, for when a browser can't display all the data. But should be additional end points.

pvretano:
From teleconference 12-FEB-2018: 
- rather than have a resultType extension perhaps this requirement could be satisfied using "microformats" with MIME types and then using the standard content negotiation mechanisms.

Issue #12, opened 2017-11-09
Response format specification: f parameter and content negotiation


jonherrmann:
The [current draft](https://cdn.rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html#_encodings_parameter_f) allows a client to pass the f-response format parameter and to use the Accept Header mechanism at the same time.

For discussion:

- Since passing different response formats is possible, I would expect an error message in that case. I'm afraid that otherwise, errors are harder to trace.
- Wouldn't the user assume that if he specifies the f-parameter, only the desired format is returned?

One possibility that would allow both mechanisms would be the use of format suffixes, like `.json / .hmtl / .xml ` :

```JSON
/buildings/{fid}:
    get:
      summary: retrieve a feature
      operationId: getFeature
      tags:
        - Features
      parameters:
        // f parameter dropped
        - $ref: '#/components/parameters/id'
      responses:
        '200':
          description: A feature.
          content:
            application/geo+json:
              schema:
                $ref: '#/components/schemas/buildingGeoJSON'
            application/gml+xml;version=3.2:
              schema:
                $ref: '#/components/schemas/buildingGML'
            text/html:
              schema:
                type: string
        default:
          description: An error occured.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/exception'
            application/xml:
              schema:
                $ref: '#/components/schemas/exception'
            text/html:
              schema:
                type: string

/buildings/{fid}.json:
    get:
      summary: retrieve a feature 
      operationId: getFeatureJson
      tags:
        - Features
      parameters:
        - $ref: '#/components/parameters/id'
      responses:
        '200':
          description: A feature.
          content:
            application/geo+json:
              schema:
                $ref: '#/components/schemas/buildingGeoJSON'
        default:
          description: An error occured.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/exception'

/buildings/{fid}.xml
...
/buildings/{fid}.html
...
```
A generic client could only use the `/buildings/{fid}` path and use content negotiation mechanism.

jonherrmann:
`Since passing different response formats is possible, I would expect an error message in that case. I'm afraid that otherwise, errors are harder to trace.`

OK, just realized that the server would respond with HTTP code 406

jvanulde:
The f-parameter should be treated like an override. This is useful in instances where you may want to override the default Accept-Header for a particular client. For example, you want to see the JSON representation in a browser which is expecting HTML.

[...]

PR #35, opened 2018-01-30
Issues 12, 22, 23, 24

### LINKED ISSUE #19: Remove OpenAPI 3.0 dependency from the Core requirements class
Discussed during the meeting on 2017-11-28:

To avoid lock-in to OpenAPI 3.0 we should change the requirements to provide a formal API description at /api and have a separate conformance class for OpenAPI 3.0, but not preclude that in the future or in parallel other versions of OpenAPI or other description mechanisms are provided by a server.

### LINKED ISSUE #24: Move XML output to extension
Having XML in as a top tier option brings in unneeded complexity. It makes a lot of sense as an extension, especially for GML 3.2 compatibility and alignment with past WFS's. 

Even though it is clearly optional in the core specification it requires more cognitive overhead to figure out what is needed, and how to take out all the references. Or we may also end up with API implementations that just leave some of the XML definitions in even though they don't support it. 

The core spec should be as short and sweet as possible, and having XML in there makes it not as short as it could be.

cportele:
The text, OpenAPI fragments and examples now focus on HTML and JSON.

Resolves #24 (in conjunction with the changes in #25/#29).


cportele:
https://rawgit.com/opengeospatial/WFS_FES/issue-19/docs/17-069.html

cportele:
Note: I mixed issue numbers, the branch should have been called "issue-24"...

cportele:
I have added the updates for issues #12, #22 and #23 to this branch/pull request as they are all somewhat related and parallel editing would have made merging later on more complicated.

cportele:
Note: openapi.yaml and swagger.yaml eventually need to be adapted to these changes.

PR #54, opened 2018-03-04
Updates for GeoJSON schema

tschaub:
I've made some changes to bring the schema for GeoJSON types closer to the spec.  I've separated these into commits that I hope are easy to interpret.

I've also published JSON schema docs for all GeoJSON object types at geojson.org in case it would be easier to refer to these:

 * [`FeatureCollection`](http://geojson.org/schema/FeatureCollection.json)
 * [`Feature`](http://geojson.org/schema/Feature.json)
 * [`GeometryCollection`](http://geojson.org/schema/GeometryCollection.json)
 * [`MultiPolygon`](http://geojson.org/schema/MultiPolygon.json)
 * [`MultiLineString`](http://geojson.org/schema/MultiLineString.json)
 * [`MultiPoint`](http://geojson.org/schema/MultiPoint.json)
 * [`Polygon`](http://geojson.org/schema/Polygon.json)
 * [`LineString`](http://geojson.org/schema/LineString.json)
 * [`Point`](http://geojson.org/schema/Point.json)


cportele:
Thanks, Tim! I consider these all as bug fixes, so I think we do not need a discussion and I can merge them directly.

I do not think we can use the JSON schemas in the OpenAPI context as OpenAPI uses its own flavour of JSON Schema, but we can discuss this in Fort Collins.

PR #68, opened 2018-03-07
Align User Guide Format with rest of spec

cmheazel:


cmheazel:
@cportele OGC Staff has tools to do this. All we need to provide is the AsciiDoc. We should have Greg schedule cron tasks to generate HTML and PDF like they do with the Testbed ERs.

cportele:
@cmheazel OK, I can also create the HTML like I create the HTML for the Core. I will do it when I do the review as I prefer to read the spec HTML. This allows us to use all the necessary plugins, too, the cron-generated D040 Testbed 14 ER is incomplete, for example.

cmheazel:
@cportele Agree with the preference for HTML.  I am concerned about the precedent.  If we insist that OGC Staff provide the transform then it will be done consistently for all documents.  It also means that there is one less tool that editors need to learn.

cmheazel:
August 22 commit. This update includes additional Information Assurance content as well as HTML versions of each section.

Issue #81, opened 2018-03-13
Duplicate reference to /req/core/fc-md-links in section 7.11 (Feature Collection Metadata)


JivanAmara:
Requirement 11 includes three bullet points, the third of which is duplicated by Requirement 13.

cportele:
This is not a duplication. In the JSON example below (from the spec, shortened), `/req/core/fc-md-items` requires that the buildings object exists in the `collections` array. `/req/core/fc-md-links` requires that the links array in the buildings object has the link with "rel" = "item".

```
{
  ..., 
  "collections": [
    {
      "name": "buildings",
      ...,
      "links": [
        { "href": "http://data.example.org/collections/buildings/items",
          "rel": "item", "type": "application/geo+json",
          "title": "Buildings" },
        ...
      ]
    }
  ]
}
```

I guess, we need to reword the requirements to make it clearer?

cportele:
I just had a closer look again at this and indeed there is a duplication. Right now we have

* Req 11, bullet 3 requires that the 'links' array has link to each collection resource. This does not belong there, the collection links belong to the 'collections' array. Delete the 3rd bullet.
* Req 12 is ok as it is (every collection available must be an item in the collections array).
* Req 13 is unclear which 'links' is meant. It needs to be clarified that this is about the 'links' in the collection, not the outer 'links' array that Req 11 references. Also the URI is a duplicate, change to `/req/core/fc-md-items-links`.

I will create a pull request to express this and then we can compare this with the existing pull request #82.

PR #82, opened 2018-03-13
Removed duplicate reference to /req/core/fc-md-links. (#81)

JivanAmara:
Resolves issue #81 , eliminating duplicate requirement for an entry in "links" for each feature collection available.


cportele:
#81 has been addressed by #97. This PR can be closed.

PR #87, opened 2018-03-14
Replace 'collection' with 'selection' in 4.1 definition of 'dataset'

### LINKED ISSUE #86: Remove the word "collection" from the definition of "dataset" to avoid confusing these terms.
4.1 "dataset" indicates all of the data served up by an entity.  Using the word "collection" in this definition confuses this with the use of "collection" to refer to sub-groupings of the dataset into feature collections.

JivanAmara:
Resolves #86


cportele:
@JivanAmara We discussed the issue and decided to implement a different approach (keep the DCAT definition as it is, but clarify the use of the term 'collection'). See #96 and the notes in #86. Let us know, if you are not happy with the approach we took. Thanks!

PR #93, opened 2018-03-21
Clarified the requirement /req/core/crs84

### LINKED ISSUE #92: Confusing crs84 requirement
The requirement ```/req/core/crs84``` is mentions the possibility for users to use other CRS than crs84.

> Unless the client explicitly requests a different coordinate reference system, all spatial geometries SHALL be in the coordinate reference system http://www.opengis.net/def/crs/OGC/1.3/CRS84 (WGS84 longitude/latitude).

It should be made clear that the Core only supports CRS84 for both the provided geometries and user provided operation parameters.

This issue is related to #65

ilkkarinne:
Clearly state that only the CRS84 is supported by the Core, fixes #92

### DIFF
--- a/core/standard/clause_7_core.adoc
+++ b/core/standard/clause_7_core.adoc
@@ -418,13 +418,14 @@ WGS84 longitude and latitude as the default coordinate reference system.
 |===
 |*Requirement {counter:req-id}* |/req/core/crs84 +
 
-For WFS 3.0 Core all geometries of both the provided features and the operation parameters provided by users 
-SHALL be in the coordinate reference system
+Unless the client explicitly requests a different coordinate reference system,
+all spatial geometries SHALL be in the coordinate reference system
 http://www.opengis.net/def/crs/OGC/1.3/CRS84 (WGS84 longitude/latitude).
 |===
 
-The Core does not allow publishing feature geometries in coordinate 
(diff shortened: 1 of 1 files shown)


cportele:
@ilkkarinne - I do not think the proposed wording works. All WFS 3.0 implementation will conform to the Core so the proposed wording would basically block any extension that would support other CRSs. The current wording is deliberate to support extensions supporting other CRSs.

ilkkarinne:
Ok @cportele: I see you point. Revised the pull to keep the req text as it was

cportele:
@ilkkarinne - Would the following text work, too?

> The implementations compliant with the Core are not required to support 
> publishing feature geometries or bounding box query parameter 
> in coordinate reference systems other than http://www.opengis.net/def/crs/OGC/1.3/CRS84. 
> That is, the Core does not specify a capability to request feature geometries in a different
> coordinate reference system. Such a capability will be specified in another part(s) of the 
> WFS 3.0 series.

The "native CRS" reference in the current PR is unclear as we do not know what the native CRS of the dataset is. It may be a projected CRS and without a mechanism to request data in a CRS that is not in CRS84 the data would not be returned in the "native CRS", but in CRS84.

Does this make sense?

ilkkarinne:
@cportele: your last proposal works for me, thanks!

cportele:
I merged it without a discussion in a SWG meeting as it only clarified the wording.

PR #110, opened 2018-04-05
Abstract Test Suite

cmheazel:
This is the Abstract Test Suite (Annex A) for the April 5 Release of WFS 3.0.


cportele:
@cmheazel - Thanks! I just had a quick look, some comments:

* In preparation for the release I had added new requirements (now 27, 28, 29) for the timeStamp, numberMatched, numberReturned parameters to resolve an open TODO. These are not yet reflected in the branch.
* Currently the ATS only covers the Core conformance class, not the other 5 conformance classes. I think we can address this by adding an issue and referencing the issue at the beginning of Annex A.

Should I make those changes and then merge the pull request?

cportele:
@cmheazel - I just saw your comment in #109. I assume my changes outlined above are ok, so I will update Annex A and then merge your pull request today or tomorrow.

PS: I plan to merge #109 on Saturday.

cmheazel:
I missed the new requirements.  It shouldn't be hard to add them.  Do you want to do it in this release or the next.  
Actually, this ATS does cover all of the conformance classes.  But they don't come into play until you are down in the weeds so they are easy to miss.  Can it be more explicit, no doubt.  Let's open this as an issue and see if we can come up with a better approach.

cportele:
I have added the three requirements and included them in a test. I have also cleaned up the AsciiDoc code and referenced the new issue #112. I saw how you added tests for GeoJSON, GML, etc., but I still think that this is incomplete. For example, these tests should then also reference the requirements from the five requirements classes and the requirements matrix should include those requirements, too. We can work on this for the next version.

I will now merge the pull request...

Issue #121, opened 2018-04-10
Behaviour of bbox and time parameters if feature does not have spatial and/or temporal properties


pvretano:
How should a wfs behave if a request has a spatial and/or temporal filter but the feature does not have spatial and/or temporal properties?  Should the wfs ignore the filter(s) or not?

cportele:
15-April-2019 Teleconference: The server will ignore the query parameters for features that have no spatial or temporal properties (as proposed by #122), i.e. they match all bbox/time parameters.

PR #122, opened 2018-04-10
Add bbox/time parameter behavior for features without bbox/time data

JivanAmara:
Likely resolution of #121.

Defines behavior of filters to match bbox/time queries if the feature doesn't have geometry/temporal data.

### DIFF
--- a/core/standard/clause_7_core.adoc
+++ b/core/standard/clause_7_core.adoc
@@ -58,6 +58,9 @@ in the response are determined by the server based on parameters of the request.
 
 A `bbox` or `time` parameter may be used to select only a subset of the features in
 the collection (the features that are located in the bounding box or time period).
+If the collection does not contain bbox data it SHALL be considered to match all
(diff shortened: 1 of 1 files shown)


core/standard/clause_7_core.adoc
```diff
@@ -58,6 +58,9 @@ in the response are determined by the server based on parameters of the request.
 
 A `bbox` or `time` parameter may be used to select only a subset of the features in
 the collection (the features that are located in the bounding box or time period).
+If the collection does not contain bbox data it SHALL be considered to match all
```
cportele: The statement is more specific, if we talk about features, not data in general. My suggestion:

"The `bbox` parameter SHALL be considered to match all features in the collection that are not associated with a location. The `time` parameter SHALL be considered to match all features in the collection that are not associated with a time stamp or period. "

cportele:
Edits transferred to #208. See  https://github.com/opengeospatial/WFS_FES/pull/208/commits/7a1e0837b5063f6cfbe4df46906eb3bf37020ce3.

PR #175, opened 2018-10-12
Change rel:item to rel:items

cholmes:
The rel link here links to the /items endpoint, but has a 'rel' of 'items'. In [STAC](https://github.com/radiantearth/stac-spec/) we extensively use rel=item to link to a specific item in our catalog. We are trying hard for WFS compatibility, and if this changes to 'items' instead of 'item' then we can be very compatible. We'd make use of rel=items to link to the wfs queryable endpoint, and rel=item to indicate a catalog link to a single item.

This was the only reference I found to this - not sure if it's even truly specified. It may be good to have some recommendations in or near the spec on the rel types that are used, and what they are used for.

### DIFF
--- a/core/standard/clause_7_core.adoc
+++ b/core/standard/clause_7_core.adoc
@@ -459,7 +459,7 @@ geometries only in the default system (WGS84 longitude/latitude).
       },
       "links": [
         { "href": "http://data.example.org/collections/buildings/items",
-          "rel": "item", "type": "application/geo+json",
+          "rel": "items", "type": "application/geo+json",
           "title": "Buildings" },
         { "href": "http://example.org/concepts/building.html",
           "rel": "describedBy", "type": "text/html",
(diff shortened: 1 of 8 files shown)


cportele:
Yes, we used `item` as it is a registered relation type, but the semantics are not clean as i is. `item` should only be used for links to a single item. I would be fine with using `items` for these links.

However, the change fixes only the example, not the following places:

* requirement 13, `/req/core/fc-md-items-links`
* the first paragraph of example 4
* the examples in Annex C

Should I take care of them?

cholmes:
Great. Sorry for missing those other places, I tried to do a search for it but obviously came up short. Would be great if you could make those changes, my time is pretty limited these days.

cportele:
OK, will do. It may take a while, but there is no rush. We will start editing the document again more actively in early 2019.

cportele:
Merge PR since the outstanding edits have been added...

Issue #176, opened 2018-10-16
Wrong edge order in BBOX description


cportele:
From Gitter:

> The OpenAPI file specifies for the spatial extent: 'West, north, east, south edges of the spatial extent.' and gives the example [-180, -90, 180, 90]. 

To be fixed (west, south, east, north).

Issue #149, opened 2018-08-21
Typos in 6.2


PeterParslow:
1. In 6.2, the third bullet point says "Geometries have to be represented in a coordinate reference system that is not based on WGS 84 longitude/latitude"

There is a missing word between "Geometries" and "have", probably "that" (or "which")

2. Similarly, the fourth bullet point says "Features have more than one geometric property", and should probably be "Features that have...."

3. In the paragraph below the bullet points, it says "as a result GML also more complex to handle"; which should probably be "as a result GML is also more....", or just "... GML is more ..."

PR #187, opened 2019-02-14
Bugfixs for issues #149 and #176

cportele:


cportele:
Merging these edits without discussion as these are bugfixes

PR #188, opened 2019-02-14
Consistent use of parameters

### LINKED ISSUE #171: Change collection name to collection id
In the specification it is defined that each collection has a `name`. This name is expected to be a unique identifier. I'd strongly vote to change this to a collection `id` due to the following reasons:
* id and title are unambiguous, name is not. Throughout software industry name is used for both titles and identifiers. Just by changing name to id the documentation would be much clearer what is expected to be added there.
* Consistency. Features have an id field, which originates from the GeoJSON spec. Why shouldn't a feature collection also have an id? It would be consistent across WFS. 

This was also shortly proposed in the comment by @cholmes here: https://github.com/opengeospatial/WFS_FES/issues/156#issuecomment-417804941

This is also discussed in other projects like STAC (https://github.com/radiantearth/stac-spec/pull/262) and openEO, which all tend to like id and title more and want to omit name because its ambiguous.

cportele:
Use the following consistently:

* /colletions/{collectionId}
* /colletions/{collectionId}/items/{featureId}
* `id` as the property in the Collection metadata

resolves #171


cportele:
Merging this as it there has been consensus in the #171 discussion.

PR #194, opened 2019-04-11
Issues with temporal data, including (open) ranges / periods #155

### LINKED ISSUE #155: Issues with temporal data, including (open) ranges / periods
I have several issues with how temporal data is handled by WFS at the moment, which I'd like to draw your attention to and hopefully facilitate some fruitful discussion:

**First issue**:  In several parts of the spec, e.g. [OpenAPI](openapi.yml) spec and [REQ_fc-time-response.adoc](core/standard/requirements/core/REQ_fc-time-response.adoc), it states that:
> temporal information is either a date-time or a period string that adheres to RFC 3339.

Examples given are:
> "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or "2018-02-12T00:00:00Z/P1M6DT12H31M12S"

RFC3339 does not specify any of that. Those formats are specified in ISO8601.

**Second issue**: Open date ranges, e.g. *2018-02-12T00:00:00Z until now*, are not possible throughout the spec, may it be for filtering or describing data. We have use cases where this is required as data updates very frequently and it is not really a solution to update the timestamps in the accommodating collection metadata every time. Also it seems useful to search for data that is just newer or older than a timestamp.

**Third issue**: The temporal filter is specified to be named `time`. I don't really like that, because it's not specified to be just a time, but a date and a time. That's counter-intuitive. I'd vote to call it `datetime` instead. This would also allow to have an extension, which would allow to just filter by time, e.g. give me all features that are available at 12:00 or between 09:00 and 12:00. There was a similar discussion on that topic by the STAC guys: https://github.com/radiantearth/stac-spec/issues/130

**Fourth issue**: In several JSON schema and OpenAPI definitions a the `format` is set to `dateTime`, but both JSON Schema and OpenAPI propose that it should be `date-time`.

cportele:
resolves #155

### DIFF
--- a/core/openapi/parameters/datetime.yaml
+++ b/core/openapi/parameters/datetime.yaml
@@ -7,7 +7,7 @@ description: >-
   * A period: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or "2018-02-12T00:00:00Z/P1M6DT12H31M12S"
 
   Only features that have a temporal property that intersects the value of
-  `time` are selected.
+  `datetime` are selected.
 
   If a feature has multiple temporal properties, it is the decision of the
   server whether only a single temporal property is used to determine
--- a/core/openapi/parameters/datetime.yaml
+++ b/core/openapi/parameters/datetime.yaml
@@ -1,10 +1,14 @@
-name: time
+name: datetime
 in: query
 description: >-
-  Either a date-time or a period string. Date and time expressions adhere to RFC 3339. Examples:
+  Either a date-time or an interval, open or closed. Date and time expressions
+  adhere to RFC 3339. Open intervals are expressed using double-dots.
+
+  Examples:
 
   * A date-time: "2018-02-12T23:20:50Z"
-  * A period: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or "2018-02-12T00:00:00Z/P1M6DT12H31M12S"
+  * A closed interval: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z"
+  * Open intervals: "2018-02-12T00:00:00Z/.." or "../2018-03-18T12:31:12Z"
 
   Only features that have a temporal property that intersects the value of
   `datetime` are selected.
--- a/core/openapi/parameters/time.yaml
+++ b/core/openapi/parameters/time.yaml
@@ -1,7 +1,7 @@
 name: time
 in: query
 description: >-
-  Either a date-time or a period string that adheres to RFC3339. Examples:
+  Either a date-time or a period string. Date and time expressions adhere to RFC 3339. Examples:
 
   * A date-time: "2018-02-12T23:20:50Z"
   * A period: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or "2018-02-12T00:00:00Z/P1M6DT12H31M12S"
--- a/core/openapi/schemas/extent.yaml
+++ b/core/openapi/schemas/extent.yaml
@@ -1,49 +1,85 @@
+description: >-
+  The extent of the features in the collection. In the Core only spatial and temporal
+  extents are specified. Extensions may add additional members to represent other
+  extents, for example, thermal or pressure ranges.
 type: object
 properties:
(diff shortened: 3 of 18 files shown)


rcoup:
Conformance checks should probably include the correct behaviour around ranges (ie. everything is >=/<=). Presumably since fractions of seconds are _optional_ in RFC3339 the ISO8601 range behaviour of truncating precision is desired?

eg. using the example range 1-2pm today UTC...

`2019-04-11T12:00:00Z/2019-04-11T12:59:59Z` equates to SQL (postgres for this example) of either
```sql
SELECT * FROM mytable
WHERE
  mytimestamp >= '2019-04-11T12:00:00Z'::timestamp
  AND date_trunc('second', mytimestamp) <= '2019-04-11T12:59:59Z'::timestamp;
```
or 
```sql
SELECT * FROM mytable
WHERE
  mytimestamp >= '2019-04-11T12:00:00Z'::timestamp
  AND mytimestamp < ('2019-04-11T12:59:59Z'::timestamp + interval '1 second')
  -- ie. mytimestamp < 2019-04-11T13:00:00Z
```
(the latter will typically perform _much_ better)

But following the ISO8601 range logic I think using the example range `2019-04-11T12:00:00.0Z/2019-04-11T12:59:59.9Z` (ie. with some arbitrary second-fraction precision) should equate to SQL of:
```sql
SELECT * FROM mytable
WHERE
  mytimestamp >= '2019-04-11T12:00:00.0Z'::timestamp
  AND mytimestamp <= '2019-04-11T12:59:59.9Z'::timestamp;
```
Which deliberately misses anything from the last 100ms in the hour. So either the user needs the query timestamp precision to exceed the system/data timestamp precision (more 9s) or doesn't specify seconds fractions at all.

(yes, this is the annoying problem with inclusive ranges, but that ship appears to have sailed)

m-mohr:
Why are inclusive ranges used? Wouldn't it be more convenient to do inclusive lower bound and exclusive upper bound?

cportele:
> Why are inclusive ranges used? Wouldn't it be more convenient to do inclusive lower bound and exclusive upper bound?

The reason is that we are referencing ISO 8601 and the intervals include the limiting time instances. See https://github.com/opengeospatial/WFS_FES/issues/155#issuecomment-422748451

m-mohr:
Hmm, okay, seems we need to live with that.

For filtering we now have open date ranges, which is great! Nevertheless, it is still not possible in the responses, e.g. the temporal extents for the collections. So instead of updating 
```json
{
  "extent":{
    "temporal":{
      "interval": [["2015-06-23T00:00:00Z", "2015-06-24T00:00:00Z"]
    }
  }
}
```
all the time, it would be great to allow:
```json
{
  "extent":{
    "temporal":{
      "interval": [["2015-06-23T00:00:00Z", null]
    }
  }
}
```

At least that was my main intention behind issue 155.

cportele:
@m-mohr - That makes sense to me. I will update the PR to allow that. 

If anyone has concerns please let me know.

cportele:
@rcoup - I have updated the test description with the text below . It does not have the details of your post, but in general the *abstract* test suites do not really go into that detail. Maybe they should. For now, it is up to the executable test suite to specify and implement the tests in detail. 

@cmheazel - please check my edits!

> Use time intervals in the tests to verify the correct behavior for time intervals. In particular check that the start and end date-time are part of the interval (i.e. everything that is `>=` than the start and `<=` the end date-time is selected). Use also fractions of seconds in the tests.

@m-mohr - I have updated the #196 to support open intervals.

rcoup:
> Presumably since fractions of seconds are optional in RFC3339 the ISO8601 range behaviour of truncating precision is desired?

Think this is a key bit... if fractions of seconds are not provided, then the filter needs to include _all values from that second_ to be compatible with ISO8601 ranges:

eg. `'2019-04-11T12:34:56.7Z'` means `<= 56.700s`
but `'2019-04-11T12:34:56Z'` means `< 57s`

cportele:
Thanks @rcoup. I will add this to the text in another commit.

PR #196, opened 2019-04-11
Extents not easily extensible as the connections to the reference systems are not clear #168

### LINKED ISSUE #168: extents not easily extensible as the connections to the reference systems are not clear
I am struggling a bit with the extent in WFS collections. Currently, it has basically fours fields:
* temporal
* spatial
* crs
* trs

To me this looks a bit dirty as the crs and trs are not closely bound to the corresponding fields. For example, one could think about adding more extents, for example temperature or another temporal extent, which (I think) happens in Meteorology. How would that work?

Example:
```
{
      "extent": {
        "spatial": [
          180,
          -56,
          -180,
          83
        ],
        "temporal": [
          "2015-06-23T00:00:00Z",
          "2015-06-24T00:00:00Z",
        ],
        "temporal2": [
          "2015-06-24T00:00:00Z",
          "2015-06-25T00:00:00Z",
        ],
        "temperature": [0, 273.15],
        "crs": "http://www.opengis.net/def/crs/OGC/1.3/CRS84",
        "trs": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian",
        "trs2": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian",
        "temperature_rs": "http://www.opengis.net/def/uom/SI/kelvin"
      }
}
```

In this case one could argue that the temporal ref system is just the same, but you probably get the point here. It is also not easy to map the fields of the ref systems to the actual fields for the values.

Also, you can't easily go through all extents because crs and trs are no extents. You can separate them by data types (array/string), but that doesn't feel right.

So I'd think there should be a consistent way to add the reference systems to the actual extents and it should be considered that there are more extents to be added to the object of extents. I think that's why the extents are actually combined and not just a separate spatial_extent and temporal extent field, right?

This could be resolved like this:
```
{
      "extent": {
        "spatial": [
          180,
          -56,
          -180,
          83
        ],
        "temporal": [
          "2015-06-23T00:00:00Z",
          "2015-06-24T00:00:00Z",
        ],
        "temporal2": [
          "2015-06-24T00:00:00Z",
          "2015-06-25T00:00:00Z",
        ],
        "temperature": [0, 273.15]
      },
      "refsys": {
        "spatial": "http://www.opengis.net/def/crs/OGC/1.3/CRS84",
        "temporal": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian",
        "temporal2": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian",
        "temperature": "http://www.opengis.net/def/uom/SI/kelvin"
      }
}
```

There are other ways to structure that and I am not saying that we should do exactly what the example shows, but it's should be more an inspiration.

cportele:
Update `extent` object based on the discussion in #168

Resolves #168

The text in Core does not have examples for the more complete cases that extensions may use. This would be how an example from the discussion in the issue would be represented:

```
{
  "extent":{
    "spatial": {
      "bbox": [[180, -56, -180, 83]],
      "crs":"http://www.opengis.net/def/crs/OGC/1.3/CRS84"
    },
    "temporal":{
      "interval": [["2015-06-23T00:00:00Z", "2015-06-24T00:00:00Z"], ["2015-06-23T00:00:00Z", "2015-06-24T00:00:00Z"]],
      "trs": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian"
    },
    "thermal":{
      "range": [0, 273.15],
      "uom": "http://www.opengis.net/def/uom/SI/kelvin"
    }
  }
}
```


core/openapi/schemas/extent.yaml
```diff
@@ -1,3 +1,90 @@
...
+          (property `interval`). The default reference system is the Gregorian calendar.
+          In the Core this is the only supported temporal reference system.
+          Extensions may support additional temporal reference systems and add
+          additional enum values.
+        type: string
+        enum:
+          - 'http://www.opengis.net/def/uom/ISO-8601/0/Gregorian'
+        default: 'http://www.opengis.net/def/uom/ISO-8601/0/Gregorian'
+
+TODO
```
m-mohr: Is that intended to be in the PR?

cportele: Oops, of course not, thanks for spotting! Lines 84 and 85 have to be removed.

PR #220, opened 2019-05-25
Remove UML and update overview

### LINKED ISSUE #217: Issues with the UML model
I had a look at the latest UML model (#209) and I have some concerns and a proposal for some modifications.

Original diagram:
image

1: 
Issue: Operations are defined on the resource that is returned by the operation.
Proposed modification: A new resource type, e.g. with name OGCAPIFeaturesConformantServer, is added, and all operations are moved to it. Appropriate parameters and return types are added to all operations.

Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) says: "A resource is not necessarily accessible via the Internet", so I would definitely argue that it is ok to add this type OGCAPIFeaturesConformantServer to the model.

2:
Issue: Operations have all the same name ("get"), a name that does not say anything about what the operation actually does or returns, and the name is technology-specific.
Proposed modification: Since all operations are moved to one resource type, they need to have different names. The new names are taken from the OpenAPI definition document in Annex B.2. E.g.: `operationId: getLandingPage`

3:
Issue: The name of a resource type should be singular.
Proposed modification: This seems an indication that something is wrong in the model. Class Collections is removed altogether. What is meant is actually an array of feature collections, see operation describeCollections.

4:
Issue: The name of a resource type should be countable, that means that mass nouns such as "metadata" and "conformance" should be avoided.
Proposed modification: This seems an indication that something is wrong in the model. Therefore, CollectionMetadata is removed altogether. Class Conformance is renamed to ConformanceClass, attribute conformanceClasses renamed to uri, the multiplicity is changed on attribute this attribute and on the end of the association from LandingPage to ConformanceClass.

5:
Issue: As far as I understand the standard, it is not possible to navigate from a feature collection directly to features, one has to go via a page. 
Proposed modification: Therefore, the navigability of that end should be set to non-navigable (visualized with a cross with the appropriate diagram settings). See https://www.uml-diagrams.org/association.html#navigability or the UML standard itself. I am aware of the fact that this is usually not done in model following ISO 19100, however, I think it could clarify this part of the standard. Note that EA seems to have a bug that does not show this cross when the relation type is aggregation. Therefore the relation type is association, but EA allows to change the field "Aggregation" to "shared" for roles also for regular associations.
image

6:
Issue: IANA link relation types, which are technology-specific, are present on the diagram as role names.
Proposed modification: Add link relation types as aliases in Enterprise Architect and apply the setting below before exporting the figure.
Another, probably better, option would have been to add those as tagged values. However, Enterprise Architect doesn't allow to visualize tagged values of attributes or operations on diagrams.
image

7:
Issue: two relations are present from Page to Page, but it should be only one relation.
Proposed modification: Remove one relation and add the roles "previousPage" and "nextPage", change navigability.

8:
Issue: The standard uses the term feature collection, which is also the preferred term in the terms and definitions.
Proposed modification: Change Collection to FeatureCollection.

New diagram:
New diagram

New UML model (extension should be changed to xml): 
[ISO 19168-1 Edition 1 - Proposal.txt](https://github.com/opengeospatial/WFS_FES/files/3187470/ISO.19168-1.Edition.1.-.Proposal.txt)

Notes:

- At first, I also added the paths as aliases, but I felt it became too cluttered. Therefore they are now present in a note. Another, probably better, option would have been to have those only as tagged values on the operations. However, Enterprise Architect doesn't allow to visualize tagged values of attributes or operations on diagrams.
- Types are taken from the ISO 19100 models where possible.
- I think it would be useful to add descriptive relation names to the model, I did an attempt to do so.

cportele:
resolves #217

There are still open questions discussed in #217, the PR is therefore marked as a work-in-progress for now.


cportele:
Conference call 10 June 2019: @cportele will update the PR based on the recent discussion in #217 and then the PR will be merged.

cportele:
Too many merge conflicts due to other edits to the requirements files. I will redo the edits in a new branch and close this PR.

PR #263, opened 2019-09-11
Clarification of wording for 3D bounding boxes 

### LINKED ISSUE #259: extent.bbox not consistently described among files
/core/openapi/ogcapi-features-1.yaml describes well the situation where there are 6 values, but /core/openapi/schemas/extent.yaml and /core/standard/clause_7_core.adoc are lagging behind.

### LINKED ISSUE #260: Weird convention for extent.bbox with 6 parameters
/core/openapi/ogcapi-features-1.yaml mentions
```
        * Lower left corner, coordinate axis 1
        * Lower left corner, coordinate axis 2
        * Lower left corner, coordinate axis 3 (optional)
        * Upper right corner, coordinate axis 1
        * Upper right corner, coordinate axis 2
        * Upper right corner, coordinate axis 3 (optional)
```
Regarding the Z axis, it seems a bit useless to take the elevation precisely at the LL and UR corners. I guess what people really want is the minimum and maximum elevation of the geometries over the whole collection.

cportele:
I have updated the textual descriptions in an attempt to reduce ambiguity:

* align wording for the bbox query parameter and the bbox member in the spatial extent of a collection;
* clarify the semantics of the values on the vertical axis.

resolves #259, resolves #260

### DIFF
--- a/core/openapi/ogcapi-features-1.yaml
+++ b/core/openapi/ogcapi-features-1.yaml
@@ -2,10 +2,10 @@ openapi: 3.0.1
 info:
   title: "Building Blocks specified in OGC API - Features - Part 1: Core"
   description: |-
-    Common components used in the 
-    [OGC draft standard "OGC API - Features - Part 1: Core"](http://docs.opengeospatial.org/DRAFTS/17-069r2.html). 
-    
-    This document is also available on 
+    Common components used in the
+    [OGC draft standard "OGC API - Features - Part 1: Core"](http://docs.opengeospatial.org/DRAFTS/17-069r2.html).
+
+    This document is also available on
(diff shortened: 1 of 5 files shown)


rcoup:
> clarify the semantics of the values on the vertical axis.

Looks good to me 👍 . 

(I personally would have gone with minimum & maximum height, but 🚲 🏠 !)

rouault:
Regarding the Z value, wondering if it is ellipsoidal height or some MSL height ? If referencing WGS84, it should theoretically be ellipsoidal height. But I'm not even sure of the semantics for GeoJSON.

cportele:
In OGC API Features it is ellipsoidal height, see http://www.opengis.net/def/crs/OGC/0/CRS84h.

Same in GeoJSON: "An OPTIONAL third-position element SHALL be the height in meters above or below the WGS 84 reference ellipsoid". See https://tools.ietf.org/html/rfc7946#section-4.

PR #279, opened 2019-10-28
New tasks

cportele:
The OGC Policies and Procedures require that ["any new work from a SWG after the primary deliverable must be approved by the TC as a new Task prior to start of that work"](http://docs.opengeospatial.org/pol/05-020r27/05-020r27.html#swg-task-process).

This is a proposal for new tasks for discussion at the SWG  meeting during the Toulouse TC meeting. After agreement by the SWG, the proposed new tasks have to be presented in the closing plenary.


ogcscotts:
The addition of the new tasks was approved by the TC in Toulouse, so the pull request can be merged and the charter updated in the OGC Portal: https://portal.opengeospatial.org/index.php?m=projects&a=view&project_id=390&tab=2&artifact_id=46634

cportele:
The charter in the portal (https://portal.opengeospatial.org/index.php?m=projects&a=view&project_id=390, requires portal access) has been updated, too. See also https://www.opengeospatial.org/projects/groups/featapiswg, which is sourced from the portal page.

@ogcscotts - I also see a Word copy of the charter in the portal. I guess we can delete this one? No need for a third copy that needs to be kept in sync.

ogcscotts:
@cportele yes, the Word document can be deleted. We have the original copy in Pending Documents if it is ever needed.

cportele:
I have moved it to a folder with older charter revisions.

PR #284, opened 2019-11-06
[WIP] add full text search capabilities

tomkralidis:
cc @pvretano 

Considerations:
- the idea behind `"wildcards": true` would be whether full text search would be a term query (`false`) searching whole words/terms, or partials
- we may want to add `wildCard`, `singleChar`, `escapeChar` as per FES 1


pvretano:
Closing as this is out of scope.  When records includes the CQL extension, it can extend and include additional requirements for full text searches including extended the filter capabilities to indicate full text search.

PR #287, opened 2019-11-06
Draft proposal for a collection summary

tschaub:
This is [a draft proposal](https://github.com/tschaub/ogcapi-features/blob/summary/proposals/002_Summary.md) for a "summary" resource that would allow clients to get additional information about items in a collection.  I've provided a bit of motivation in the proposal doc itself.  This is a very "drafty" draft - plenty of room for discussion on alternative ways to get this done.

I didn't see the [Alternative Schema](https://github.com/opengeospatial/ogcapi-features/blob/master/proposals/001_Alternative-Schema-Proposal.md) until after creating this PR (and then reshaped this to use the same proposal template).  Interesting that that tries to address some of the same concerns.  Good evidence that there is motivation.

### DIFF
--- a/proposals/002_Summary.md
+++ b/proposals/002_Summary.md
@@ -1,0 +1,111 @@
+# Collection Summary
+
+## Metadata
+
+| Tag | Value |
+| ---- | ---------------- |
+| Proposal | [Summary](https://github.com/tschaub/ogcapi-features/tree/summary/proposals/002_Summary.md) |
+| Authors | Various |
+| Review Manager | TBD |
+| Status | Draft |
+| Implementations | None |
+| Issues | None |
+| Previous Revisions | None |
+
+## Introduction
+
+This proposal suggests that the spec define a `summary` resource (all names TBD) that describes a collection.  It may be that this information belongs in the existing collection info (e.g. at `/collections/{id}`), but for the purposes of this doc, we'll say it is going to be a new resource (e.g. at `/collections/{id}/summary`).
+
+## Motivation
+
+There are few motivations for this:
+
+ * A client should have a way to get the formal schema for items in a collection if one exists.  The current suggestion to use a link with `rel="describedBy"` doesn't fully address this need since it doesn't provide a robust way for a client to get the schema for multiple content types (for example, if a service supports GML, KML, and GeoJSON output, the `type` and `rel` of a link is not enough information for a client to know which schema applies to which output format).
+
+ * A "filter" extension should have a way to describe which properties are filterable and how clients should access a specific property in a filter expression.
+
+ * A "search" extension should have a way to describe which properties can be used in sorting.
+
+ * An "editing" extension should have a way to describe which properties are writeable (e.g. in an update) or might want to allow for a different schema for each method (e.g. one schema for create, one for update).
+
+## Proposed solution
+
+## Example
+
+Here is an example of what this summary resource might look like for a "buildings" collection (e.g. at https://example.com/collections/buildings/summary)
+
+```json
+{
+  "properties": {
+    "application/geo+json": {
+      "height": {
+        "schema": {
+          "href": "http://example.com/collections/buildings/schema#/definitions/height"
+        }
+      },
+      "age": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
+        },
+        "filter": {
+          "disallowed": true
+        },
+        "search": {
+          "unsortable": true
+        }
+      },
+      "taxID": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
+        },
+        "editing": {
+          "readonly": true
+        }
+      }
+    }
+  }
+}
+```
+
+Notes on the above example:
+
+ * The properties of the `properties` object (ack!) are content types.  Using an object means that there will be one member per content type.  So it looks like the responses `content` object in OpenAPI 3.
+
+ * The `schema` property would be defined by the core Features spec.  In the case of JSON schema, an href could deeply link to the spceific definition.  In the case of an XSD doc, there could be two properties: `href` and `path` (maybe?) so an XPath expression could be provided to refer to the part of the schema specific to a single property.
+
+ * Individual extensions could add their own members to each of the properties object.  For example, the "filter" expression could define a `filter` object.
+ 
+ * I like the idea that defaults are `false` - so the absence of a property implies the default property.  This means that by default, a property would be sortable for example.  If a service wanted to disallow sorting on a particular property, they would add `"unsortable": true` to change the default behavior.
+
+## Detailed design
+
+TODO
+
+## Backwards compatibility
+
+This proposal doesn't introduce any breaking changes.
+
+## Alternatives considered
+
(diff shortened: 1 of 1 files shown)


extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+        "filter": {
+          "disallowed": true
+        },
+        "search": {
+          "unsortable": true
+        }
+      },
+      "taxID": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
```
rouault:
```suggestion
          "href": "http://example.com/collections/foo/schema#/definitions/taxId"
```

extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+        "schema": {
+          "href": "http://example.com/collections/buildings/schema#/definitions/height"
+        }
+      },
+      "age": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
+        },
+        "filter": {
+          "disallowed": true
```
rouault:
```suggestion
          "allowed": false
```
Negative property names are hard to understand because of the double negation

extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+        },
+        "search": {
+          "unsortable": true
```
rouault:
```suggestion
          "sortable": false
```

extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+        }
+      },
+      "taxID": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
+        },
+        "editing": {
+          "readonly": true
```
rouault: Assuming that most propeties would be writable for a editing ccontext, perhaps:

```suggestion
          "writable": false
```

extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+
+Notes on the above example:
+
+ * The properties of the `properties` object (ack!) are content types.  Using an object means that there will be one member per content type.  So it looks like the responses `content` object in OpenAPI 3.
+
+ * The `schema` property would be defined by the core Features spec.  In the case of JSON schema, an href could deeply link to the spceific definition.  In the case of an XSD doc, there could be two properties: `href` and `path` (maybe?) so an XPath expression could be provided to refer to the part of the schema specific to a single property.
+
+ * Individual extensions could add their own members to each of the properties object.  For example, the "filter" expression could define a `filter` object.
+ 
+ * I like the idea that defaults are `false` - so the absence of a property implies the default property.  This means that by default, a property would be sortable for example.  If a service wanted to disallow sorting on a particular property, they would add `"unsortable": true` to change the default behavior.
```
rouault: ok I let the above comments without seing this rationale. To ponder if we prefer the consistency of missing property meaning false, or if we want to avoid negative property names

extensions/summary/summary.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Alternatives
+
+In discussion with others here at the Features/STAC sprint, we discussed a few other alternatives.  This isn't a comprehensive set of notes, but we covered things like the list below:
+
+### A new resource per extension
+
+We considered the alternative that each extension would add its own resource at a new path.  For example, the filter extension might add a `/collections/buildings/queryables` resource.  Then the search extension might add a `/collections/buildings/sortables` resource.  And then the editing extension might add a `/collections/buildings/writeables` resource.  This began to feel unscalable and awkward for clients who want to get a complete picture of how to work with a collection.
+
```
rouault: and those various capabilities would still have a need to point to the schema, so a lot of redundancy

proposals/002_Summary.md
```diff
@@ -0,0 +1,111 @@
...
+        "filter": {
+          "disallowed": true
+        },
+        "search": {
+          "unsortable": true
+        }
+      },
+      "taxID": {
+        "schema": {
+          "href": "http://example.com/collections/foo/schema#/definitions/age"
```
rouault:
```suggestion
          "href": "http://example.com/collections/foo/schema#/definitions/taxId"
```

extensions/summary/summary.md
tschaub: Good catch.  I used `taxID` actually so either need to change the anchor to `/#definitions/taxID` or change the property name to `taxId`.

extensions/summary/summary.md
tschaub: I agree.  And am open to alternatives.  I think there is some value in having the default be the zero value though.  If we have an `allowed` property and it is absent, it would be nice if that were the same as `false`.  This means that the default would be to have no properties be filterable.

Again, I can drop this and go with the idea that if `allowed` is absent, the default is allowed (but that is also kind of weird).

extensions/summary/summary.md
tschaub: Happy to switch all of these.  But see above for the consistency I was trying for: if a boolean property is absent, it implies `false`.  I can tell this is an uphill battle though.

extensions/summary/summary.md
tschaub: I agree about the awkwardness of the double negative.  And I think this wouldn't be something that people would adhere too.  So I'll switch them.

proposals/002_Summary.md
tschaub: I changed the property name to `taxId` (lowercase `d`) in 5a0f9d83042d06af9bf46d29321d2a4fdcf123c7.

extensions/summary/summary.md
tschaub: I removed this point in 1084d463427da4935d353eb0929cdb655dc06ce7.

extensions/summary/summary.md
tschaub: Addressed in 5a0f9d83042d06af9bf46d29321d2a4fdcf123c7.

proposals/002_Summary.md
```diff
@@ -0,0 +1,109 @@
...
+  "definitions": {
+    "age": {
+      "type": "number",
+      "filterable": false,
+      "sortable": false
+    }
+  }
+}
+
+Even if this were possible, our concern was that generic schema parsers would not expose these additional properties to a client.
```
m-mohr: This is possible. JSON schema allows for custom keywords and many validators allow to implement custom logic for this.

proposals/002_Summary.md
```diff
@@ -0,0 +1,109 @@
...
+ 
+## Detailed design
+
+TODO
+
+## Backwards compatibility
+
+This proposal doesn't introduce any breaking changes.
+
+## Alternatives considered
```
m-mohr: Have you had a look at the STAC Collection summaries? Even if you not consider it, it would be good if names would not conflict too much...

cportele:
Code Sprint 2024-02-27: A [draft of "Part 5: Schemas"](https://docs.ogc.org/DRAFTS/23-058.html) exists and addresses the need for feature schemas to describe the feature properties ("returnables"), the editable properties ("receivables"), the properties that can be used in filters ("queryables") and the properties that can be used to sort features ("sortables"). The part is expected to be submitted for OGC Architecture Board Review and then Public Comment soon. We are, therefore, closing this draft PR and invite comments on the draft of Part 5.

PR #297, opened 2019-11-22
QGIS added to implementations

elpaso:


elpaso:
@cportele  what kind of links do you want me to add? PR ? Code? Manual/documentation?

cportele:
@elpaso - I was thinking manual/documentation/examples/blog post so that readers quickly get an idea of the scope of the OGC API implementation. If the are no such links that's fine, too. Thanks.

elpaso:
I cannot comment on the client implementation (CC: @rouault), for the server we have the documentation in place at: https://docs.qgis.org/testing/en/docs/user_manual/working_with_ogc/server/services.html#wfs3-ogc-api-features
There is also a demo server in our CI pipeline:  http://138.201.120.72:8084/qgisserver_demo_wfs3/wfs3/

cportele:
Thanks @elpaso - I have updated it with the info and will merge it into master. If anything should be changed (or added for the client - @rouault), let me know.

PR #298, opened 2019-11-22
It's filter-lang right?

aaime:
@pvretano at the spring we agreed on filter-lang I believe?


pvretano:
Yes, I have implemented filter-lang as we discussed at the OAPIF/STAC sprint.

aaime:
Cool. This PR updates the README accordingly.

Issue #299, opened 2019-11-25
License recommendation in Core


heidivanparys:
There seems to be a mistake in the recommendation regarding the license: the recommendation says

> [...] include an item for each supported encoding [...]

but that should be a link to a license.

cportele:
Meeting 2019-11-25:  The updated wording is fine and probably easier to read. The change is editorial. We will keep issues and PRs like this open until a future update of the standard and tag them accordingly. Such a future update may be bugfixes or editorial aligment with ISO 19168-1 or OGC API Common. 

Note: The text of the recommendation is correct as it is. "the `links` property should include an item" is the same as "the `links` property should include a link" - the value of the `links` property is an array.

PR #300, opened 2019-11-25
Update license recommendation

heidivanparys:
1. Describe "license" instead of "encoding".
2. Use "media type" instead of "content type", as in the rest of the document.


heidivanparys:
This is a proposal for solution of #299.

Issue #301, opened 2019-11-25
Core: requirements class vs. conformance class


heidivanparys:
I noticed that the dependencies of certain requirements classes link to other requirements classes but the text of the links says "Conformance class" instead "Requirements class".

An example:

![image](https://user-images.githubusercontent.com/11427611/69529183-0dc52880-0f70-11ea-9ee1-232f641ad239.png)

cportele:
Meeting 2019-11-25:  Yes, "Requirements class" would have been correct, but this choice was deliberate as many readers understand "conformance class", but not "requirements class" (unless they are experts of the OGC modular specification policy). Since there is a 1-to-1 relationsship it seemed ok to be intentionally imprecise and use "conformance class" for both to have more intuitive text for the non-OGC-experts. But it is not optimal as "requirements class" is used sometimes, too, including in the table heading. This is a general topic and it would be good to have a consistent approach across OGC specs. @cmheazel will work on a proposal.

heidivanparys:
@cportele @cmheazel 

Wouldn't the consistent approach across OGC specs be to follow the OGC modular specification policy? 🙃

As the OGC modular specification policy is being used as input in the revision of 19105, it would be advantageous if the aforementioned proposal could be ready before the ISO 19105 EC meeting in Japan (9th and 11th of December) so it can be taken into account there.

Regarding "to have more intuitive text for the non-OGC-experts": An explanation regarding the way OGC standards are structured could be added to to the guide to the OGC API - Features in a section called "Introduction to the OGC modular specification". Such a section could also contain figures, e.g. figures like the ones proposed by Denmark in the comments on ISO/CD 19105, if those are accepted.

cportele:
@heidivanparys - The document follows and is consistent with the mod spec policy! Req classes have dependencies to req classes and everything is mirrored 1-to-1 in conformance classes. The only thing is that for simplicity we sometimes write "conformance class X" and mean the "requirements class that is associated with conformance class X".

In practice most spec readers do not read a spec from start to end. They just jump to the specific part they are interested in at the moment. At least this is what I often do and probably most others, too. Most questions that I get on the spec could easily be answered by reading the spec 😏. And most readers are almost certainly not interested in the metamodel underpinning the standards (maybe beside those that design standards), so I think it is important to write standards so readers do not have to be familiar with the ISO Directives or the OGC policies.

[...]

PR #302, opened 2019-11-25
Change "conformance class" to "requirements class"

heidivanparys:
Change "conformance class" to "requirements class" where needed.


heidivanparys:
This is a proposed solution for #301.

Issue #303, opened 2019-11-25
Core: description of status code 406


heidivanparys:
A server could also return status code 406 Not Acceptable when the language negotiation (mentioned in 7.10. String internationalization) or the content negotiation based on `Accept-Charset` or `Accept-Encoding` fails.

I would like to see a more general description of 406.

F.eks.:

Content negotiation failed. For example, the `Accept` header submitted in the request did not support any of the media types supported by the server for the requested resource.

cportele:
Meeting 2019-11-25:  Agreed. This is something that should be addressed in OGC API Common (once we reference Common, the table will likely be removed from the Features document).

cmheazel:
@heidivanparys API Common has been updated using your text.

PR #304, opened 2019-11-25
Update description for status code 406

heidivanparys:
A server could also return status code 406 Not Acceptable when the language negotiation (mentioned in 7.10. String internationalization) or the content negotiation based on `Accept-Charset` or `Accept-Encoding` fails.


heidivanparys:
Proposed solution for #303.

Issue #306, opened 2019-11-25
Core: example 4


heidivanparys:
A suggestion for a minor editorial change: example 4 could be updated to have an explanation regarding all links present.

![image](https://user-images.githubusercontent.com/11427611/69555148-01a68e80-0fa3-11ea-84c0-55016c4d046f.png)

cportele:
Meeting 2019-11-25:  Same as #303.

cmheazel:
This change has been made to API-Common

PR #307, opened 2019-11-25
Update example 4

heidivanparys:
Add explanation regarding link relation type "self" and "enclosure".


heidivanparys:
This is a proposal for #306 .

PR #310, opened 2019-11-26
Created mappings btwn 17-069r3 sections & spec element URIs

ghobona:
The mappings in this commit are intended to help configure OGC servers to make the URIs of specification elements resolvable.


cportele:
@ghobona @ogcscotts @pvretano @cmheazel 

See my [commit](https://github.com/opengeospatial/ogcapi-features/pull/310/commits/86913f65a16677262f787064836096df8a76af65) for more information about the changes to the mappings.

As part of my review I noticed a few inconsistencies in the published document:

* The test cases use `.../ats/{confClassId}/{testCaseId}` while the conformance classes use `.../conf/{confClassId}`. I checked the rules for specification elements and noticed that we should have used `.../conf/{confClassId}/{testCaseId}` for the test cases. 
* For the OAS 3.0 conformance class we use `oas3` for the id, but everywhere else we use `oas30` (requirements classes, normative statements, test cases). 
* I noticed four test cases which are not included in the current spec document (two are mentioned, but not included), for now their URIs would resolve to the test case in the GitHub repository

My proposal is to fix all these issues in 1.0.1 (whenever we will work on it, i.e. keep them for now as errata) and then continue to support the then deprecated URIs (e.g. the `ats` and `oas3` URIs). We would then also redirect the URIs of the four missing test cases when the test cases have anchors in the document.

Comments?

cportele:
Meeting 2019-12-23: Merge this PR, ask @ghobona to update the definitions server. @cportele to create a new issue for the corrigendum input.

ghobona:
@cportele 

I have submitted a request to the OGC Infrastructure Office for the mappings to be updated on the server.

cportele:
@ghobona - Is there any estimate when the update will be live?

ghobona:
@cportele

There are a few errors in the changes you have suggested.

* The mappings have to be to HTML fragments in the generated HTML document and not the Asciidoc source on Github. Notice that the base of the other URLs is http://docs.opengeospatial.org/is/17-069r3/17-069r3.html

* There is no HTML fragment with the identifier ATS_fc-md-extent or ats_core_fc-md-extent in the HTML markup. So I suspect that the specific test suite is missing from the standard.

* There is no HTML fragment with the identifier ATS_fc-md-items-links or ats_core_fc-md-items-links in the HTML markup. So I suspect that the specific test suite is missing from the standard.

* There is no HTML fragment with the identifier ATS_fc-md-items or ats_core_fc-md-items in the HTML markup. There is however an href link pointing to such an identifier (see attached screenshot), so I suspect that the specific test suite is missing from the standard.

* There is no HTML fragment with the identifier ATS_fc-md-links or ats_core_fc-md-links in the HTML markup. There is however a href pointing to such an identifier(see attached screenshot), so I suspect that the specific test suite is missing from the standard.

I have copied in @kstegemoller , to keep him in the loop.

Screen Shot 2020-02-05 at 09 16 00

cportele:
@ghobona - Yes, see my comment above:

> I noticed four test cases which are not included in the current spec document (two are mentioned, but not included), for now their URIs would resolve to the test case in the GitHub repository

So, if I understand it correctly, we have to remove the four entries until we publish version 1.0.1. I have made the change in https://github.com/opengeospatial/ogcapi-features/commit/828cf50f4657d8edbeb69b30059eea4d77082559.

PR #352, opened 2020-03-20
Fix links to RFCs

vmx:
The links to the RFCs had a typo.


vmx:
Wow, that was quick, thanks a lot!

PR #392, opened 2020-05-24
Fix issue-336.

pvretano:
This pull request fixes issue 336 but it also fixes an issue with the definition of the "interval" production that arose during validation.  Apparently an empty string (i.e. "") is not valid according to the https://www.icosaedro.it/bnf_chk/bnf_chk-on-line.html BNF validator.


cportele:
looks good to me

PR #442, opened 2020-10-12
Add a table that crosswalks each resource endpoint with each HTTP met…

pvretano:
…hod to indicate for which combinations behaviour is defined in this standard.


cportele:
Meeting 2020-10-12: Review until the next meeting and merge then, if there are no comments.

PR #443, opened 2020-10-12
Change order of ENVELOPE coordinates to match those of bbox in part 1.

pvretano:


cportele:
Meeting 2020-10-12: Review until the next meeting and merge then, if there are no comments.

PR #455, opened 2020-11-22
update Queryables to use JSON Schema

### LINKED ISSUE #383: queryables link relation type
cc @alpha-beta-soup

If one wants to provide a link to a given collection's `queryables` endpoint, what link relation should they use (if any)?

Is there a suitable IANA type (`describes`?) Should there be an extension (`ogc-cql-queryables`)?

### LINKED ISSUE #424: Harmonize the "queryables" resource with OGC API Styles
See http://docs.opengeospatial.org/DRAFTS/20-009.html#rc_queryables. Both resources have the same role. There should only be a single requirements class for queryables.

cportele:
updates Queryables resource, closes #424

adds link relation type for queryables, closes #383


cportele:
Meeting 2020-11-23: Keep open for review until 2020-12-07 (next meeting at Member meeting) and then resolve comments or merge.

Issue #451, opened 2020-10-30
Future work items (Part 6++)


cportele:
The current work items of the OGC Features API working group are progressing well. At the same time there are questions or requests for additional capabilities. Since the OGC processes take time to get approval for the working group to work on new topics, it is time to identify and prepare the topics that should be addressed next. This issue is intended to discuss, identify and agree on the most important additional capabilities that we should work on next. 

We need to present the identified topics to the OGC members to start the process, typically in a plenary (the next one would be in early December).

As a reminder, here is what the group charter says:

> Standards are important for interoperability. At the same time, it is important that standards only state requirements that are important for a significantly large group of users. Proposals for new parts of OGC API Features or change requests to existing parts must identify the user group that will benefit from the proposal and include the commitment for three independent implementations for each proposed conformance class; otherwise the proposal will be considered out-of-scope.

In addition, we of course need to have at least one editor for each proposal to edit the document and drive the process.

Proposals for new tasks will only be considered, if these criteria are met.

Many extensions may be useful beyond Features and we should ask for input from members of the other OGC API working groups, too.

cportele:
To start the discussion, here is a list of topics that I have seen requested most over the last months:

* Schemas: We have identified three different types of schemas for the features of a feature collection that we need and need to describe in a consistent way (at least in cases where the dataset has a schema) - the queryable properties of the features that may be used in filter expressions, the returnable properties included in feature representations retrieved from the API and the storage properties that are processed in requests mutating a feature. We already need the queryables schema for part 3 (so that clients know, what they can filter on) and the storage schema for part 4 (so that clients know, what data they have to provide for new or updated features), but it really should be a separate document, at least eventually. The returnables schema is important for clients that depend on schemas (many GIS clients do), currently they have to guess a schema for a subset of the instance data.
* Property selection: One or more query parameters to reduce the properties that are returned for each feature. This has been requested frequently to reduce the data volume that has to be transferred. For example, dropping the geometry from polygon data where only a table needs to be presented to a user can dramatically reduce the size of the response and the response time.
* Geometry simplification: A query parameter to indicate a scale or zoom level at which the features will be displayed. The returned curve/surface/solid geometries will be simplified accordingly. The capability reduces the size of the response and the transmission time plus the client can directly use the geometry without additional processing.
[...]

PR #485, opened 2020-12-14
draft charter update (new tasks, general update)

cportele:
See #451. In addition to the new proposed tasks, the document has been updated where updated information is available.

@pvretano - Please have look before I upload it to Pending in the OGC portal.


pvretano (commented):
Clemens, this looks good to me.

PR #490, opened 2020-12-16
Issue 487

pvretano:
Remove the extensions/cql/examples/examples.txt file since it used an old syntax for both JSON and CQL.  
Add a bunch of example taken from examples.txt and updated to use the new JSON and CQL syntax.


extensions/cql/standard/schema/examples/example02.txt
```diff
@@ -1,3 +1,3 @@
-EXAMPLE 2: Any item where eo:instrument starts with OLI 
+EXAMPLE: Find any item where the instrument name starts with 'OLI'.
 
 filter=eo:instrument like 'OLI*'
```
cportele, line 3: Shouldn't this be something like `filter=eo:instrument like '^OLI'`, if we use regex?

extensions/cql/standard/schema/examples/example07.txt
```diff
@@ -1,3 +1,3 @@
-EXAMPLE 07: Comparison and spatial predicates in a single expression
+EXAMPLE: An example of comparison and spatial predicates in a single expression.
 
-filter=eo_instruments LIKE 'OLI*' AND footprint INTERSECTS POLYGON((43.5845 -79.5442, 43.6079 -79.4893, 43.5677 -79.4632, 43.6129 -79.3925, 43.6223 -79.3238, 43.6576 -79.3163, 43.7945 -79.1178, 43.8144 -79.1542, 43.8555 -79.1714, 43.7509 -79.6390, 43.5845 -79.5442))
+eo_instruments LIKE 'OLI*' AND footprint INTERSECTS POLYGON((43.5845 -79.5442, 43.6079 -79.4893, 43.5677 -79.4632, 43.6129 -79.3925, 43.6223 -79.3238, 43.6576 -79.3163, 43.7945 -79.1178, 43.8144 -79.1542, 43.8555 -79.1714, 43.7509 -79.6390, 43.5845 -79.5442))
```
cportele, line 3: Same comment on the LIKE as above.

Also, why has `filter=` been removed here, but not in other examples? Is there a pattern?

extensions/cql/standard/schema/examples/example11.json
```diff
@@ -0,0 +1 @@
+{ "like": [{"property": "owner"},"*Jones*"] }
```
cportele, line 1: Again, same comment on the pattern. This applies also to the other examples.

extensions/cql/standard/schema/examples/example20.txt
```diff
@@ -0,0 +1,3 @@
+EXAMPLE: Find building features where the building was built before 2015.
+
+built BEFORE '2015-01-01'
```
cportele, line 3: Should be without quotes, ie. `built BEFORE 2015-01-01`.

This applies to other cql-text examples below as well.

cportele (changes requested):
I fixed a few obvious errors, but there are a few discussion topics:

* In the text examples sometimes there is `filter=` and sometimes not. Probably remove it everywhere?
* It seems like you are using "*" as a wildcard in LIKE, but that is not a posix regex. Or am I missing something?
* Temporal literals are not in quotes or should we allow both quoted and unquoted?

jratike80 (approved):
I did not find logical errors: bounding boxes have min<max, geometries seem to be valid, datatypes (strings, numbers, boolean) are used correctly. I do not master regexp (who does?) and can't say if ^ is needed in some places.

There may be something to check with DURING. Should the times be separated with "/"? Also in http://docs.ogc.org/DRAFTS/19-079.html#enhanced-temporal-operators, shouldn't the time in JSON example be enclosed between double, not single quotes, like in the example22.json in this PR?

pvretano:
@jratike80 Yes, "^" is needed in several places as well as a couple of other REGEX changes; always get SQL pattern matching semantics mixed up with REGEX. ;)  Yes, there should be a "/" in the DURING and that has been fixed.  No, there should be no single or double quotes around temporal literals in the CQL TEXT encoding of DURING.  It should simply be:
```
updated DURING 2017-06-10T07:30:00Z/2017-06-11T10:30:00Z
```
... fixed that too.  Hey, thanks a lot for taking the time to review the examples.  I really appreciate it!

jratike80:
Example 16 has also this **JSON** example:

```
{
   "during": [{"property": "touchdown"},['1969-07-16T13:32:00Z',
                                         '1969-07-24T16:50:35Z']]
}
```

jratike80 (commented):
Shouldn't there still be the wildcard part "OLI followed by any character, any count"  

`LIKE '^OLI.*'`?

PR #494, opened 2020-12-22
Issue 489

pvretano:
Go back to using SQL-like LIKE semantics as per the discussion in Issue 489.


pvretano:
@cportele review and if it looks good, go ahead and merge.

PR #521, opened 2021-03-17
Moved the Compliance/Product Database text to the main README file

ghobona:
Greater visibility of Certified Compliant Products is consistent with direction from the PC. So this Pull Request moved the section about the Compliance/Product Database to the front.


cportele (approved):
This looks good to me.

PR #524, opened 2021-03-21
Issue-515: Add optimistic locking conformance class

pvretano:


cportele (approved):
This looks good. I will merge this, so the HTML is updated before the SWG meeting.

I have a few editorial comments:

- The `Last-Modified` values in the examples are not consistent (the value after the first PUT is the same at the beginning). 
- The definition of optimistic locking needs to be split into the definition (following the rules of definitions in ISO) and an additional note.

I will fix these in the master branch after merging, also the chapter numbering of the adoc files.

Other comments that we can discuss later:

- Since `Last-Modified` is used in the examples, maybe we should mention it in the text somewhere, even if we do not add any normative text related to the header?
- The `If-Match` requirements are only requirements for the server (our standardisation target) and what the server has to do with requests. Maybe we can make that clearer in the wording and also clarify explicitly in the informative text that the server will send a 412 in case of a missing `If-Match` header. We may also add more guidance to clients in informative text.
- Shouldn't we also allow "*" as a value in `If-Match` (if the server wants to support this)? Currently that is not allowed.

PR #535, opened 2021-03-23
Update nls-fi implementation page

jampukka:
Change links from "beta" to "avoin" ("open" in finnish). Add information about required api-key.


cportele:
Thanks @jampukka - Do you also want to state the supported requirements classes in https://github.com/opengeospatial/ogcapi-features/blob/master/implementations/README.adoc or do you prefer to leave the entry as it is?

jampukka:
Sure, I added what we currently have ready. `html` and of Part 3 should get finished in the upcoming months if I have some time.

PR #537, opened 2021-03-23
Reference ArcGIS Runtime APIs

hmassih:


hmassih:
Hi @cportele, could you please review this PR.

cportele (approved):
Thanks @hmassih, this looks good. I tried it myself based on the example with the Java Runtime and it worked well using the example as a starting point.

Would it make sense to add a screenshot for each example to give an idea how the end result should look like? Here is one I have created:

Screenshot 2021-03-24 at 15 42 54

hmassih:
Thank you @cportele for reviewing this PR. Of course adding screenshots makes perfect sense, I’ll be adding them early next week.

pvretano (approved):

hmassih:
@cportele @pvretano thank you for approving this PR - I've added two screenshots per your recommendation, please have a look.

Issue #562, opened 2021-04-19
CQL: Naming of operators


heidivanparys:
As the naming is not yet decided 100% according to https://github.com/opengeospatial/ogcapi-features/issues/549#issuecomment-817877359: personally I would prefer the following, for consistency...:

- prefix "s" for all **s**patial operators
- prefix "t" for all **t**emporal operators
- prefix "a" for all **a**rray operators

cportele:
2021-05-10: Use consistently prefixes "s_", "a_" and "t_" for spatial, array and temporal operators.

cportele:
PR #587 addresses the temporal operators

aaime:
Given the language is deviating from the CQL as defined in the CSW extension, I would suggest to also change the name of the extension to avoid confusion. "CQL2"? "API CQL?"? Maybe "AQL" (API Query Language).

When GeoServer added extra behavior to the original CQL, we dubbed the result ECQL (Extended CQL). However, those were simple extensions and a valid CQL statement was also a valid ECQL. In this case the deviation is backwards incompatible.
Oh, I know... "NoCQL"? Just kidding!!

Issue #532, opened 2021-03-22
CQL: Spatial predicates supporting WKT geometry are too complex for "Simple CQL"


jerstlouis:
I would like to see Spatial predicates and Temporal predicates in their own conformance class, not in "Simple CQL", as it currently groups together essential thing like property/value filters, with much less essential things such as clients describing and intersecting geometry in WKT.

It would make more sense to me for intersects to be grouped with that 'advanced spatial', as if services implement intersects then they likely can implement everything else easily. But if the server doesn't have a use case for WKT geometry provided by clients, it is a lot to implement that is all tied in together with the simple filters & logic.

While "Core" supports filters, those are subject to dataset attributes conflicting with defined query parameters, and generally speaking a `filter` query parameter feels like the more correct way to do filtering. The filtering also allows logical operators which Core does not support.

cportele:
Meeting 2021-06-07: Since there were no objections, we will follow this approach.

cportele:
PR #587 addresses the temporal operators

Issue #483, opened 2020-12-12
CQL: Restrict timezone in temporal literals to UTC ("Z")?


cportele:
OGC API Features uses RFC 3339 for timestamps, i.e., all timestamps have a time zone.

Currently temporal literals in CQL can be in any time zone. Should at least Simple CQL be restricted to UTC?

heidivanparys:
This is perhaps related to this? The OAB discussed this apparently before (I don't have access to that part of the OGC portal), but some others here perhaps do) https://github.com/opengeospatial/geopackage/issues/530#issuecomment-724845716

cportele:
I do not have access to the OAB part of the OGC portal either (access to the OAB project worked until about 2 months ago and that changed without information about the change). But since nothing is included in the [Policy Directives](https://portal.ogc.org/public_ogc/directives/directives.php), either the OAB is still discussing something or there was no result. If an OAB member has any further information, please let us know. 

cc @ogcscotts

ogcscotts:
OAB Issues should now be visible to everyone, but only if you have a link to the Issue. We are trying to refine the access permissions for the OAB as well as update some operations (including considering moving Issues to GitHub). Stay tuned - hope to make the access simpler very soon.

cportele:
Meeting 2021-06-21: We plan to fix the timezone to URC ("Z"), because it makes life easier in evaluating filter expressions. This would also be consistent with the GPKG approach and also the next generation 911 services use UTC. If there are any concerns about such a decision, please add a comment here in the next 2 weeks.

cportele:
Meeting 2021-07-05: No comments have been added, so we will fix the timezone to "Z".

PR #587, opened 2021-06-10
Update temporal operators

### LINKED ISSUE #533: CQL: Consider renaming anyinteracts to tintersects
If there is no strong requirement to leave CQL unchanged, consider renaming `anyinteracts` to `tintersects`, as the current name does not seem to carry its meaning of intersecting time, and will just confuse everyone.

### LINKED ISSUE #565: CQL: instant-interval and instant-instant relations
Regarding the following in 9.3. Requirements Class "Enhanced Temporal Operators":

> CQL supports date and timestamps as time instants, but even the smallest "instant" has a duration and can also be evaluated as an interval. For the purposes of determining the temporal relationship between two temporal expressions, an instant is treated as the interval from the beginning to the end of the instant. The following diagram illustrates the meaning of the temporal operators.

Yes, time is continuous, which means that any moment of time can be divided into an infinite number of smaller moments. But, time is _measured_ by dividing time into time intervals with a specific duration (temporal resolution = granularity). In practice, a time interval whose duration is less than the temporal resolution used is _considered to be_ an instant.

So to take example 15: when using a temporal resolution of one second, 1969-07-20T20:17:40Z is treated as an [instant](https://www.w3.org/TR/owl-time/#time:Instant) (although we know that we can subdivide it in fractions of seconds, but that is not the resolution used in this case) and 1969-07-16T13:32:00Z/1969-07-24T16:50:35Z is an interval.

The [Time Ontology](https://www.w3.org/TR/owl-time/), listed as a reference, also makes a distinction between instants and time intervals.

As a minimum, the "TOVERLAPS" and "OVERLAPPEDBY" relations are only meaningful for interval-interval relations, not for instant-interval or instant-instant relations. So how should a server react when the client submits a query containing `1969-07-20T20:17:40Z TOVERLAPS 1969-07-16T13:32:00Z/1969-07-24T16:50:35Z`? That query is not meaningful. Should an error code be returned?

How about https://www.w3.org/TR/owl-time/#time:hasBeginning and https://www.w3.org/TR/owl-time/#time:hasEnd? They have time:Instant as range and are conceptually similar to https://www.w3.org/TR/owl-time/#time:intervalStarts and https://www.w3.org/TR/owl-time/#time:intervalFinishes. BEGINS is mapped to time:intervalStarts, but a time interval has a start instant.

### LINKED ISSUE #570: CQL: usefulness of the current Simple CQL regarding temporal predicates
I am wondering how useful the current Simple CQL requirements class is regarding temporal predicates, as the only relationship supported is `ANYINTERACTS`. I understand that implementing all relationships from Allen's interval algebra is too much, but, as also argued in #565, I think that a distinction should be made between

- instant-instant relations
- instant-interval relations
- interval-interval relations

I would like to suggest including instant-instant relations in Simple CQL, i.e. supporting before, equals and after for instants. Isn't it the case that any other relation involving intervals actually can be expressed by comparing their start and end instants?

cportele:
* closes #565 (except the rebranding to CQL2 -> open a new issue when merging this PR)
* closes #570
* closes #533 
* addresses #562 (only temporal operators)
* addresses #532 (only temporal operators)


cportele:
@heidivanparys - please have a look at the changes

pvretano (approved):

pvretano:
@cportele looks good.  Merge if you like ...

cportele:
@pvretano - thanks, let's see, if @heidivanparys has any comments on the changes

heidivanparys:
@cportele 
Has there been a misunderstanding regarding #570?

I added a commit to illustrate what I meant with "I would like to suggest including instant-instant relations in Simple CQL, i.e. supporting before, equals and after for instants.": https://github.com/heidivanparys/ogcapi-features/commit/cd6a6fa9d064a79b34492056990b3b4b0751de94

Would that be possible, especially if #483 would be accepted (only instants expressed in UTC allowed)? That way, behaviour would be similar to that of relational databases, where <, = and > can be used to compare instants, allowing to form more complex expressions on the client side.

cportele:
> Has there been a misunderstanding regarding #570?

I think/hope what is in the PR is what is documented as the approach documented during the meting (see https://github.com/opengeospatial/ogcapi-features/issues/565#issuecomment-826884331), but we can discuss.

Also note that it was decided to move anyinteracts (or t_intersects) from Simple CQL to the other temporal operators, see #532.

cportele:
Meeting 2021-06-14: We should include a discussion of date or date-time handling in Simple CQL when we discuss the comparison operators like `<` or `>` (including in the ATS). This should work with SQL backends, but will it create issues for NoSQL backends (MongoDB, Couchbase, ElasticSearch, GQL, GraphQL, etc). @pvretano will investigate. If others would like that other backends should be considered, let us know here. Note there does not have to be a one-to-one mapping, there will always be parsing and processing involved.

Issue #546, opened 2021-04-02
Part 3: Temporal literals in the text encoding


pvretano:
Temporal literal might be problematic to parse as defined.  Without looking at the queryables and conformance classes one might think that 1969-07-16 is an arithmetic expression equalling 1946.  Similarly 1969-07-16/1969-07-24 could also be interpreted as the numeric expression 1930.99187404774.

We should probably consider what SQL does ... e.g. date('1969-07-16').  That is a builtin function with a date string.

As I mentioned, you can figure out the context of the value by looking at the queryables and conformance classes but that seems like a heavy lift just to parse a date.

cportele:
Just to note that such a change would break existing CQL implementations. If we make changes like these, then we should reconsider other changes that we have discarded because of the existing CQL grammar (and also change the name to differentiate the CSW CQL and the new CQL).

cportele:
2021-05-10: For implementations that support arithmetic expressions, we need a delimiter to disambiguate a temporal literal from an arithmetic expression. Question is whether it is sufficient to make it a string or if we need an explicit function/cast like in SQL? Discuss this again in the next meeting. Check also what other languages are using, but if a string is sufficient we should keep it simple.

cportele:
Meeting 2021-06-07: If the context is always clear, then a wrapper function like `toDate()` would be uncessary and a quoted literal would be sufficient. @pvretano will analyze this and make a proposal.

pvretano:
Did a survey of how popular databases handle dates.

- In all cases date literals are string literals.
- In many cases, the format of the string is ISO8601.
- All databases have functions to convert a string literal to a date.  

Here is a brief survey of some of the popular databases:

- Oracle and Postgres use TO_DATE('... date string...').
- MySQL uses STR_TO_DATE('...date string...').
- Casandra, Elastic search and SQL Server use CAST('... date string...' as date).
- MongoDB uses DATE('... date string...')

In most cases these function take other arguments too but I don't think they are relevant to CQL since they are mostly pattern arguments so that the date string does not need to be in ISO8601 format.  That is not the case for CQL.

Assuming that we start with the proposition that we need, at the very least, to change bare temporal literals to string literals then, there are numerous places in the current CQL grammar where it would be difficult to disambiguate a string literal from a date literal without referring to an external schema.  This makes parsing a CQL expression a little inconvenient and so, I think we should go for the least ambiguous and simplest to parse approach and change the definition of `instant` from this:
```
[...]

Issue #541, opened 2021-03-31
Part3: LIKE is unnecessarily complex


jampukka:
I wonder if we could make the LIKE predicate simpler than what it is now in Part3 draft by removing the ability to change wildcard characters (WILDCHAR, SINGLECHAR, ESCAPECHAR) from the default values. There's very little benefit in having this ability at all or at least I'm having hard time seeing it.

I do agree we should have a case-insensitive version of LIKE in simple-cql. However I wonder if having a separate predicate for this (ILIKE for example) would be easier to implement?

aaime:
Agree with @jampukka

pvretano:
I mostly agree with simplifying LIKE.  However, I think ILIKE does not intuitively imply case less LIKE so I would prefer we just stick with the current setup.

cportele:
I would prefer ILIKE, because that seems to be the way this is done in a number of SQL implementations.

aaime:
By the way, I've found that LIKE's "NOCASE" keyword is also used in the "in" predicate. Seems again a bit complex, and also somewhat inconsistent, in that a "in" is just a shorthand for a chain of or-ed equality comparisons. But in an equality comparison, there is no way to express case insensivity. Should I open a separate issue for it?

pvretano:
@cportele I don't like it there either ... but I am not religious about it ... and Oracle SQL does not define an ILIKE using UPPER() or LOWER() instead.
General questions for everyone ...
(1) Do we wanted case insensitive comparisons in core (i.e. simple CQL)?  I think yes.
(2) How should we do it consistently everywhere.  I say forget NOCASE and ILIKE and simply define a built-in function named UPPER() and/or LOWER() that people can inject anywhere they want case-less comparisons.  Any other proposals?

cportele:
Meeting 2021-04-12: Current thinking:

- Remove options to change the WILDCHAR, SINGLECHAR, ESCAPECHAR values and keep them fixed.
- Case insensitive should be supported in Simple CQL
- Instead of (NO)CASE the definition of functions UPPER/LOWER enables also case insensitive comparisons for `=`etc, so this seems to be most flexible and consistent with SQL implementations.

jampukka:
[...]

Issue #489, opened 2020-12-16
CQL: Question about using % as a wildcard in LIKE


jratike80:
**Requirement 18, /req/simple-cql/like-predicate**

> The patternExpression SHALL be a POSIX regular expression (see https://pubs.opengroup.org/onlinepubs/9699919799/).

In the examples https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/examples/examples.txt the SQL-alike percent syntax is used

`CQL: owner LIKE 'Mike%'`

This percent syntax is defined at least in https://portal.ogc.org/files/?artifact_id=831 "Catalog Interface
Implementation Specification (Version1.0)"

```
<character pattern> ::= <character string literal> /* In a character pattern
the character percent is used as a wildcard to represent an
arbitrary string. This allows LIKE to implement the effect of many
characters matching operations, such as: contains, begins with,
ends with, not contains, not begins with, not ends with, and so
forth. For example:
 attribute like '%contains_this%'
 attribute like 'begins_with_this%'
 attribute like '%ends_with_this'
 attribute like ‘d_ve’ will match ‘dave’ or “dove””
 attribute not like '%will_not_contain_this%'
 attribute not like 'will_not_begin_with_this%'
 attribute not like '%will_not_end_with_this’*/
```

I must admit that I am not so good in reading the Posix regexp document.  If patternExpression SHALL be a POSIX regular expression, does it mean that the percent syntax is not valid in LIKE?

pvretano:
@jratike80 you are looking at the **WRONG** examples file.  The file cql/examples/example.txt is an OLD file using an OLD syntax.  I was going to remove it from the repo but instead have decided to move all the examples in example.txt into separate files in cql/standard/schema/examples.  All examples from example.txt will be updated to use the correct syntax.  Unfortunately you are reviewing FASTER that I can edit.  Please stand by and I will tag you to review the PULL request that updates the examples.

pvretano:
@jratike80 I have removed the old examples.txt file and instead added a bunch of examples to extensions/cql/standard/schema/examples from that file which have also been updated to the correct CQL and JSON syntax.  Please take a look and let me know if you need any problems before we merge.  The link to the PR is: https://github.com/opengeospatial/ogcapi-features/pull/490.

jratike80:
This is clear now. Follow-up questions and a comment:

- Are clients supposed to have full support for POSIX regexp?
[...]

PR #592, opened 2021-06-12
Simplify the LIKE operator

pvretano:
- Simplify the LIKE operator by removing the WILDCARD, SINGLECHAR, ESCAPECHAR, and NOCASE qualifiers.
- Add the builtin functions UPPER() and LOWER() to allow caseless comparisons.


pvretano:
Question:  Right now I have simplified the LIKE expression quite a bit.  I only define the wildcard character (%) and the escape character (\).  Is this too simple?  Show we let the RHS of the LIKE operator be a regex? ... maybe regex is too complex?

cportele (commented):
Thanks @pvretano. Two questions for discussion:

* We did discuss regex before, but decided against it, see #489. However, I thought that we would keep "_" as the single char wildcard since this is what LIKE does. Did we decide against it? I agree, the general wildcard "%" is more important, but I think that users of LIKE would expect support for "_", too?
* UPPER/LOWER are now special cases in the BNF grammar, not two pre-defined functions. Wouldn't pre-defined functions be cleaner?

pvretano:
- @cportele oh, I thought we decided to strip it right down BUT I agree the single char wildcard would be useful.  I can't tell what character you are proposing for the single char match.  It looks like double quotes on my screen but I would propose we use the underbar "_".
- I'm not sure what you mean by "pre-defined" in this context.  Do you mean that the function are not built into the grammar the way the spatial operators are and instead we simply document that two functions need to be implemented named upper/lower and described their prototypes using the syntax defined in the grammar?

cportele:
> I can't tell what character you are proposing for the single char match. It looks like double quotes on my screen but I would propose we use the underbar "_".

Oops, yes, I meant the underscore char. I forgot this is for italics in markdown...

> I'm not sure what you mean by "pre-defined" in this context. Do you mean that the functions are not built into the grammar the way the spatial operators are and instead we simply document that two functions need to be implemented named upper/lower and described their prototypes using the syntax defined in the grammar?

Yes. But I also see the point that the spatial operators are also functions that we have in the grammar. I think we also need to consider how we will handle #546 in the BNF in this context. Let's discuss that in the meeting today or next week.

cportele:
See https://github.com/opengeospatial/ogcapi-features/issues/541#issuecomment-860830350

PR #594, opened 2021-06-13
Move all spatial operators into the spatial_operators conformance class.

pvretano:
Previously, the SIMPLE_CQL conformance class included the INTERSECTS spatial operator.  This operator has now been moved into the ENHANCED_SPATIAL_OPERATORS conformance class which was renamed to simply SPATIAL_OPERATORS conformance class.


cportele (approved):
Looks good

PR #599, opened 2021-07-01
consistent use of height / ellipsoidal height

### LINKED ISSUE #598: height vs. ellipsoidal height
Comments from Roger Lott / IOGP received during the ISO 19168-2 process:

There are several references to 'height' for which the context is a geometric, i.e. ellipsoidal height associated with a geographic 3D CRS rather than a physical height associated with a vertical CRS and which is the typical understanding of height when its type is unspecified. For the avoidance of doubt it would be more complete if these occurrences were changed from 'height' to 'ellipsoidal height'. Specific occurrences:

* 6.2.1, last line: Change "...with height" to "...with ellipsoidal height".
* 6.3.1, Req. 8: Change "...with height" to "...with ellipsoidal height".
* 6.3.2, Req. 12: There is minor inconsistency in treatment of the 2D CRS84 in the document. Req. 8 has "... for coordinates without height" but here in Req 12 has "... for coordinates without ellipsoidal height" . Improve document by making these consistent. For references to a 2D CRS as in Requirement 12 then it is slightly better to use 'height', so change "...with ellipsoidal height" to "...with height".
* 6.3.3.2, paragraph beginning "GeoJSON normatively supports ...": Change "(without height: CRS84; with height: CRS84h)" to (without height: CRS84; with ellipsoidal height: CRS84h).

In addition, in Annex A.1 / Test 2 the comma in the test method after i.e. is superfluous. Change "i.e.," to "i.e".

I will create a PR for the changes.

cportele:
closes #598

### DIFF
--- a/core/openapi/parameters/bbox.yaml
+++ b/core/openapi/parameters/bbox.yaml
@@ -16,8 +16,8 @@ description: |-
   WGS 84 longitude/latitude (http://www.opengis.net/def/crs/OGC/1.3/CRS84)
   unless a different coordinate reference system is specified in a parameter `bbox-crs`.
 
-  If the value consists of six numbers, the coordinate reference system is
-  WGS 84 longitude/latitude/height (http://www.opengis.net/def/crs/OGC/0/CRS84h)
+  If the value consists of six numbers, the coordinate reference system is WGS 84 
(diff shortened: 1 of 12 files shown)


pvretano:
@cportele looked over the changes.   They look good to me.

cportele:
Meeting 2021-07-05: @cportele to ensure that this is also handled correctly in Part 1, then merge.

Issue #603, opened 2021-07-05
CQL -> CQL2


cportele:
While the grammar design continues to be consistent with and based on the underspecified CQL in CSW, there have been a number of changes and it should be clear that CQL as specified in Part 3 has breaking changes compared to the old version. Consistent with the pattern that uses "WKT2" for the current WKT for CRS specification we will use "CQL2" as the short form for this version of the Common Query Language.

cportele:
Meeting 2021-07-19: Keep "Common Query Language" and use "CQL2" as the short form. Add a note to distinguish it from Cassandra Query Language and other CQLs that may be out there. Try to be clear and use "OGC CQL2" in general communication.

Issue #579, opened 2021-05-12
Consider splitting up Simple CQL into multiple conformance classes


philvarner:
In the context of STAC Items, the biggest piece missing from the Part 1 "items" endpoint is a *formalized* mechanism for filtering based on fields in the Feature/Item's Properties. Implementations can support arbitrary query parameters to support this, but there's no "queryables" mechanism to discover what fields can be filtered on. For example, a common filter criteria is cloud cover, and there's no well-defined way to express a simple `cloud_cover <= 10`. An implementer may desire to only support this one predicate, and is satisfied with using the existing `datetime` and `bbox` for spatial and temporal filtering. 

However, with the current Simple CQL conformance class, an implementer would be forced to support more than just simple comparison operations (e.g., gt, lt, etc.) and would also need to support all the logical operators and the more complex comparison operators, IN, LIKE, BETWEEN and IS NULL. (They would not *actually* need to implement INTERSECTS or ANYINTERACTS if they don't advertise any queryables of spatial or temporal types) This makes conforming to the CQL spec much more difficult that the needs of the implementation require, and will likely lead to implementers doing something proprietary or, even worse, only partially implementing CQL but advertising that it is conformant!

Additionally, I don't like the use of "Simple" for the class, as it is still a pretty complex language. I think a word like "basic" would be more appropriate here. Also, this class has "CQL" in the name, whereas as none of the other "operator" classes do.

I'd like to propose to divide Simple CQL into these separate conformance classes:

- Basic Comparison Operators - `=,<,<=. >, >=`
- Enhanced Comparison Operators - `IN, LIKE, BETWEEN, IS NULL`
- Logical Operators - `and, or, not` (these could also go into a "basic logical and comparison operators" class, but it feels a little cleaner to split them out)
- Basic Spatial Operators - `intersects` 
- Basic Temporal Operators - `anyinteracts`

While this does introduce some overhead, in that now an implementation has to advertise 5 classes instead of 1, I think there is a large benefit in allowing implementations with limited needs to properly implement a subset of these conformance classes, instead of simply not conforming to CQL at all. I think that this will result in both more robust implementations and provide an "on-ramp" for an implemenation to support more and more of the CQL conformance classes over time.

Thanks for your consideration of this!

pvretano:
[...]

PR #610, opened 2021-07-19
Implement issues marked as "to be drafted"

### LINKED ISSUE #483: CQL: Restrict timezone in temporal literals to UTC ("Z")?

### LINKED ISSUE #493: The syntax for spatial predicates is difference than that for other predicates.  Why?
The syntax for most predicates in CQL is "expression OPERATOR expression".  The syntax for the spatial operators, however, is OPERATOR(expression,expression).  This seem oddly incongruous and was inherited from CSW which probably inherited it from OGC 06-103r4.  OGC 01-103r4 was probably using a functional notation (i.e. OPERATOR(expression,expression)=0|1) to accommodate the limitations (i.e. no Boolean type) of RDBMSs at the time.

So, is this something we should change? 

I should point out that unlike 06-103r4, the requirement in CQL is that OPERATOR(spatialExpr,spatialExpr) evaluates to TRUE or FALSE so someone writing a CQL filter does not need to say **filter=INTERSECTS(geom,POLYGON(...)=1**.

### LINKED ISSUE #546: Part 3: Temporal literals in the text encoding

### LINKED ISSUE #581: Part 3: duplicate $comment attributes give linting error in the CQL JSON Schema
The duplicate $comment attributes give linting error when using spectral in the CQL JSON Schema. This also gives me a warning in VS Code. 

This can be resolved by making all the $comment fields unique. 

```
spectral lint cql.json --ruleset .spectral.yml
```

using .spectral.yml containing:

```
extends: "spectral:oas"
rules:
  oas3-api-servers: off
  oas3-schema: warn
  tag-description: true
  oas3-parameter-description: true
```

### LINKED ISSUE #607: CQL JSON: "not equal" binary comparison not available
The `not equal` predicate is available for CQL Text in the BNF

```bnf
...

comparisonOperator = eq | neq | lt | gt | lteq | gteq;

neq = lt gt;

...
```

but not for CQL JSON encoding:

```json
// ...

    "binaryComparisonPredicate": {
         "oneOf": [
            { "$ref": "#/$defs/eqExpression" },
            { "$ref": "#/$defs/ltExpression" },
            { "$ref": "#/$defs/gtExpression" },
            { "$ref": "#/$defs/lteExpression" },
            { "$ref": "#/$defs/gteExpression" }
         ]
      },

// ...
```

I understand that you could always do a `not` with an `eq` inside, but this is needlessly complex and the same would be true for CQL Text. So this `property <> 123` could be written as `NOT property = 123`

My suggestion would be to add an `neExpression` (or `neqExpression`) and add it as a possibility to the `binaryComparisonPredicate`.

cportele:
closes #483
closes #546
closes #581
closes #493
closes #607


cportele:
Note: I kept #579 and #603 separate, because those changes should be reviewed separately.

cportele:
Meeting 2021-07-19: Agreed to merge.

PR #614, opened 2021-08-02
Issue 602: Using binary comparison operators to evaluate simple temporal expressions using time instants.

pvretano:
Even though this is evident in the BNF, add a permission (and examples) that states that instead of using temporal operators, binary comparison operators may be used evaluate simple temporal predicates involving time instants.


cportele (approved):
LGTM

PR #615, opened 2021-08-02
Issue-578: Patch error in JSON divExpression and make consistent use of arithmetic and comparison operators.

pvretano:
- Patch error in JSON divExpression.
- Consistently use +,-,/,*,<,>,<=,>=,<> rather than a mix of symbols and FORTRAN-like two-character codes.


cportele (approved):
Using the same operator tokens in Text and JSON makes sense to me.

PR #618, opened 2021-08-10
reorganise conformance classes

### LINKED ISSUE #579: Consider splitting up Simple CQL into multiple conformance classes

cportele:
partially addresses #579


extensions/cql/standard/clause_2_conformance.adoc
```diff
@@ -53,25 +56,39 @@ The specific set of operators defined in this requirements class is:
...
 ** less than
 ** less than or equal to
 ** greater than
 ** greater than or equal to
-** like
 ** is null
-** between
-** in
+
+All expressions on the left hand side of the operator must be a property, all expressions on the right hand side must be a literal value.
```
cportele, line 66: Meeting 2021-08-16: introduce "property-literal" terminology to make the connection to property-property and literal-literal more clear.

PR #619, opened 2021-08-10
Use CQL2 as the acronym

### LINKED ISSUE #603: CQL -> CQL2

cportele:
Changes "CQL" to "CQL2", "cql" to "cql2". Adds explanation in the introduction.

To avoid merging conflicts, this PR builds on PR #618.

closes #603


cportele:
Meeting 2021-08-16: Good to merge.

cportele:
Closing because of merge conflicts, replacing with #620

PR #622, opened 2021-08-30
issue-557: consistent handling of NULL in predicate evaluation in the requirements

pvretano:
Add a NOTE indicating that a predicate that evaluates to NULL is the same as a predicate that evaluates to FALSE.  We do this so that we do not, in each requirement, have to deal with the case where the predicate evaluates to NULL.


cportele:
Meeting 2021-08-30: Also remove the mentions of "true, false or null", e.g. in the Abstract or in section 8.2, so that we discuss this only once. @pvretano will make the change and then merge the PR.

PR #649, opened 2021-11-08
review requirements class dependencies

### LINKED ISSUE #638: Review and clarify requirements (dependencies between operators and other requirements classes)
Raised by @jampukka on Gitter:

> What about services that would like to implement "Advanced Comparison Operators" (LIKE and IS NULL aren't available via Basic CQL2) but don't support "Spatial Operators" and/or "Temporal Operators"? Permission 5 (`/per/advanced-comparison-operators/in-predicate) currently allows one to implement "Advanced Comparison Operators" without implementing "Property-Property Comparisons" but there seems to be no Permission that would allow for not supporting spatialLiterals and/or intervalLiterals in inListOperand.
>
> Contrasting this to BETWEEN which is currently limited to numericExpressions only it feels a bit weird that IN allows spatialLiterals and intervalLiterals.

Intervals only have to supported, if Temporal Operators is supported, etc. We should review and clarify dependencies. 

From a practical point of view, I am not sure there is much value in using intervals or spatial geometries with IN.

cportele:
* remove interval/spatial geometry from IN
* add OGC API Features Part 3 to bibliography
* correct references to "filter" requirements class
* remove normative references to OGC API Features Part 3
* update LIKE operator to remove support for numeric/boolean/instant literals
* remove constructs from cql2.yml that are not supported by the OpenAPI 3.0 schema object
* some editorial improvements

closes #638

### DIFF
--- a/cql2/standard/annex_bibliography.adoc
+++ b/cql2/standard/annex_bibliography.adoc
@@ -8,3 +8,4 @@
 * [[iso8601-1]] ISO 8601-1:2019, **Date and time — Representations for information interchange — Part 1: Basic rules**
 * [[iso8601-2]] ISO 8601-2:2019, **Date and time — Representations for information interchange — Part 2: Extensions**
 * [[iso15836-2]] ISO 15836-2:2019, **Information and documentation — The Dublin Core metadata element set — Part 2: DCMI Properties and classes**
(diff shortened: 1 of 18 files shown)


cportele:
Meeting 2021-11-08: @cportele will update the PR to remove arithmetic expressions from the characterExpression rule of LIKE. After that he will merge the PR so that @pvretano can update his PR.

PR #655, opened 2021-11-16
Issue 601: Revised JSON encoding using op/args keys.

pvretano:


pvretano:
@cportele can you validate the OpenAPI 3.0 version before merging?  I checked the JSON examples using https://www.jsonschemavalidator.net/ and they all seem to validate.

pvretano:
06-DEC-2021: Decided to merge now to minimize merge conflicts.  Any errors that we find after the fact we can log as bugs to be fixed. @cportele has the action to review over the holidays! ;)

PR #663, opened 2022-01-03
clarify behaviour of 'limit'

### LINKED ISSUE #368: Clarification on limit parameter ...
Consider that a server has specified that the min and max values for the limit parameter are 10000.  If a client then makes a request for 50000 records should the server throw an exception (since 50000 is greater than 10000) or should it simply return 10000 records?

### LINKED ISSUE #640: in OAFeat Part 1, phrasing of the semantics of the limit parameter
It would be good if the `limit` parameter were explicitly described as being allowed to be larger than the advertised maximum value without returning an error code.

In the OpenAPI spec, the limit param is described as:

```
The optional limit parameter limits the number of items that are presented in the response document.
```

In the text spec, this is said:

```
The `limit` parameter may be used to control the subset of the selected features that should be returned in the response, the page size.
```

One thing that is not clearly stated is what the behavior should be if the `limit` parameter is larger than the advertised or default maximum value.  There is an example in the text:

```
If you ask for 50000, you might get up to 10000 (server-limited) and if there are more, a next link;
```

which indicates that larger values than maximum should not result in an error, but instead should treat the parameter as if it were the maximum, but it would be good to have this explicitly stated in the definition of the limit parameter.

cportele:
closes #368
closes #640

### DIFF
--- a/core/standard/abstract_tests/core/ATS_fc-limit-response.adoc
+++ b/core/standard/abstract_tests/core/ATS_fc-limit-response.adoc
@@ -7,4 +7,5 @@
 ^|Test Method |. Count the Features which are on the first level of the collection. Any nested objects contained within the explicitly requested items  are not be counted.
 . Verify that this count is not greater than the value specified by the `limit` parameter. 
 . If the API definition specifies a maximum value for `limit` parameter, verify that the count does not exceed this maximum value.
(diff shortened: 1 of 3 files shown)


cportele:
Meeting 2022-01-17: Agreed to merge after @cportele fixes a typo.

PR #668, opened 2022-01-31
follow-up edits from PR #641

cportele:
There were still some outstanding comments/edits from the discussion in PR #641:

* Add explanatory text (https://github.com/opengeospatial/ogcapi-features/pull/641#issuecomment-988080226)
* Removed definition of "collation" - it was not used in the text and it is not necessary to introduce the term (https://github.com/opengeospatial/ogcapi-features/pull/641#discussion_r785835148)
* Add "unicode normalization" as a term (https://github.com/opengeospatial/ogcapi-features/pull/641#discussion_r785839217)
* Add "unicode case folding" as a term (https://github.com/opengeospatial/ogcapi-features/pull/641#discussion_r785842023)
* Correct examples (https://github.com/opengeospatial/ogcapi-features/pull/641#discussion_r795699292)

### DIFF
--- a/cql2/standard/clause_4_terms_and_definitions.adoc
+++ b/cql2/standard/clause_4_terms_and_definitions.adoc
@@ -7,11 +7,6 @@ This document also uses terms defined in the OGC Standard for Modular specificat
 
 For the purposes of this document, the following additional terms and definitions apply.
 
-[[collation-def]]
-==== collation
-a set of rules that indicate how to compare and sort character string data
-the process of ordering units of textual information https://www.unicode.org/glossary/#collation[Glossary of Unicode Terms]
-
 [[collection-def]]
 ==== collection
 a body of **resources** that belong or are used together; an aggregate, set, or group of related **resources** (http://docs.opengeospatial.org/DRAFTS/20-024.html#terms_and_definitions[OGC 20-024, OGC API - Common - Part 2: Collections]).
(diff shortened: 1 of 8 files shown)


cportele:
Another question is whether it should be "arguments" in the JSON representation of a function or "args" (like for the operators).

pvretano (approved):

pvretano:
@cportele looks good ... go ahead and merge if you like.

pvretano:
@cportele about "arguments" or "args", we should be consistent across the board I think.  Besides, "args" is shorter!

cportele:
OK, I have changed to "args" everywhere.

cportele:
@pvretano - Before merging, there is still the following note below requirement 9 that needs to be resolved:

> The references in D need to be verified. Also need to have some discussion about NFC (cacnonically-composed form) or NFD (canonically decomposed form). I think the correct thing is to say that ACCENTI() must do NFC …​ but I need to verify.

pvretano:
Hmm ... why do we need to say anything about NFD or NFC?  As long as one or the other is applied consistently in the implementation then code points should be comparable ... no?

cportele:
I guess so, but it was you who wrote that note :)

pvretano:
@cportele yeah because I was not sure but on further reflection I'm having second thoughts about getting into that level of detail.  Perhaps we can discuss at today's meeting.

PR #684, opened 2022-04-11
Open issues for Part 3 and CQL2

### LINKED ISSUE #617: Add Abstract Test Suite
* [x] Part 3
* [x] CQL2

### LINKED ISSUE #672: CQL: give CASEI and ACCENTI distinct conformance classes
In the current standard, it seems one cannot implement only CASEI and advertise the "http://www.opengis.net/spec/cql2/1.0/req/case-insensitive-comparison", as that seems to include also the ACCENTI function.

I find this setup onerous from the implementation point of view, while case insensitive comparisons are easily implemented in most data sources, and are already implemented by anyone that dealt with Filter Encoding, accent insensitive requires an ad-hoc setup in the data source, and sometimes even language-specific, e.g., see:
https://www.postgresql.org/docs/12/collation.html

May I suggest giving the two functions separate conformance classes?

### LINKED ISSUE #674: CQL2: interval() as temporal literal
I am updating our implementation to CQL2 and I have run into the following problem:

All temporal intervals are defined as "temporal literals", however, this is incorrect as the rule is not restricted to literals:

```text
temporalLiteral = instantLiteral | intervalLiteral;

intervalLiteral = "INTERVAL" 
                  leftParen instantParameter comma instantParameter rightParen;

instantParameter = dateInstantString
                 | timestampInstantString
                 | quote dotdot quote
                 | propertyName
                 | function;
```

For example, I can define a temporal "literal" value as `INTERVAL(start, end)`, referencing two properties.  This does not look like a literal to me. It is more a regular function that can take a literal or a property reference (or another function) - similar to the pre-defined functions CASEI or ACCENTI.

One option to address this could be to define INTERVAL as a proper function, like CASEI and ACCENTI. DATE and TIMESTAMP could stay as qualifiers for strings that represent instants. This would update the `function` and `temporalLiteral` rules as follows:

```text
function = identifier leftParen {argumentList} rightParen
         | "CASEI" leftParen characterExpression rightParen
         | "ACCENTI" leftParen characterExpression rightParen
         | "INTERVAL" leftParen instantParameter comma instantParameter rightParen;

temporalLiteral = dateInstant | timestampInstant;
```

However, this would have an impact on the JSON encoding where we would get

```json
{ "name": "interval", "args": [ "2019-09-09", ".." ] }
```

instead of

```json
{ "interval": [ "2019-09-09", ".." ] }
```

In case of property references we would get:

```json
{ "name": "interval", "args": [ { "property": "start" }, { "property": "end" } ] }
```

Thoughts?

### LINKED ISSUE #683: The arithmeticExpression production in CQL2 is not quite right ...
The production does not seem to enforce precedence rules correctly and does not, for example, support an expression like `1 + 5 + 9`.  It also does not support exponents.  I think the production should look something like this:
````
arithmeticExpression = arithmeticExpression + arithmeticTerm
                     | arithmeticExpression - arithmeticTerm
                     | arithmeticTerm;
arithmeticTerm = arithmeticTerm * powerTerm
               | arithmeticTerm / powerTerm
               | powerTerm;
powerTerm = arithmeticFactor caret powerTerm
          | arithmeticFactor;
arithmeticFactor = leftParen arithmeticExpression rightParen
                 | arithmeticOperand;
arithmeticOperand = numericLiteral
                  | propertyName
                  | function;
````

cportele:
Part 3:

* [x] add ATS, closes #617

CQL2:

* [x] add ATS, closes #617
* [x] distinct requirements classes for CASEI and ACCENTI, closes #672
* [x] add clarifications on data types
* [x] update grammar and JSON Schema, closes #674
* [x] update arithmetic expression, closes #683


cportele:
This pull request addresses all open issues.

PR #698, opened 2022-06-06
Various clarifications as per issue-515.

pvretano:
To address comments by @cportele in issue 515.


cportele:
Changes now included in #774

PR #710, opened 2022-06-27
Adding missing `^` operator and comma to 7.11

jerstlouis:
The list of operators was missing the power (caret) operator and a comma after `i.e.`.


jerstlouis:
Actually, was the intent for exponents to _not_ be supported? It's also missing from the conformance section, and I see now that the `caret` rule is not defined anywhere in the BNF, even though it is used within `powerTerm`.

cportele (approved):

cportele:
Meeting 2022-07-04: We need three additional changes before merging this:

* Add `caret` rule to BNF
* Update operator list in comment above powerTerm
* Same addition, but in clause 2 (conformance)

@cportele will update and merge afterwards.

Issue #723, opened 2022-07-01
CQL2 as an expression language (not only boolean predicates)


jerstlouis:
CQL2 would actually be very useful as a general expression language, if we would only remove the concept that it needs to evaluates to a boolean expression. Such an expression language is needed in several OGC specifications, including for speciying derived fields (e.g., `properties=ndvi:(B5-B4)/(B5+B4)` in an extension for _OGC API - Coverages_ and _OGC API - DGGS_), for specifying parameter values in _SymCore_, as well as for _OGC API - Features - Part 5: Search/Queries_ extension.

Also if I understand the grammar correctly, currently e.g., `active` (referencing a boolean queryable called `active`) would not be a valid CQL expression, even though it does evaluates to a boolean expression, while `active = true` would be. Even in the context of  predicate/filter, e.g. in a styling rule selector, it would be nice to be able to use the shorter syntax, e.g. `[active] { visible = true }`. Related to point 2 of #705.

If the requirement that an expression evaluates to a boolean was in the _OGC API - Features - Part 3: Filter_ extension instead, that would leave CQL2 ready to use as a generic expression language.

pvretano:
@jerstlouis about the fact that CQL expression evaluate to a boolean expression I am not opposed to removing that or at least qualifying it a little further (e.g. when used in a filter expression, CQL expression SHALL evaluate to true or false) but that is something that will need to be discussed in the SWG.
About a boolean queryables, that's easy enough to add and I'm not opposed to that either.

cportele:
I think my main concern is that most of the CQL2 language is about boolean predicates (all the logical, comparison, spatial, temporal, array predicates are boolean-valued). Only the functions, arithmetic expressions and the operators like CASEI are not. I.e., the main mechanism for extending the language to result in non-boolean results would be additional operators or functions, but that is probably ok. 

But if we open the scope to a general spatio-temporal expression language, it would seem a little odd that the language does not include more operators to, just for example, construct geometries (buffer, union, intersection, etc.) or other non-boolean values. 

Could it be an option to simply explicitly open the possibility to extend the CQL2 grammar / JSON schema in an extension to non-boolean results (and start working on this). Such an extension should then also define key operators for the main usage patterns outside of filter expressions. I guess such an extended language should then also have a different name to be clear about the scope?

jerstlouis:
@cportele 

[...]

PR #732, opened 2022-07-05
add note about future evolution to an expression language

cportele:
See https://github.com/opengeospatial/ogcapi-features/issues/723#issuecomment-1173929247

I have added the text as a note to the scope section since the statement is about the (future) scope of the document.

@pvretano @jerstlouis

### DIFF
--- a/cql2/standard/clause_1_scope.adoc
+++ b/cql2/standard/clause_1_scope.adoc
@@ -6,3 +6,10 @@ This document defines
 
 * A text encoding for a CQL2 filter;
 * A JSON encoding for a CQL2 filter.
+
+NOTE: The CQL2 grammar contains many language elements of a general purpose, spatio-temporal expression language.
+The focus of this version of the language are boolean-valued filter expressions. It is planned to investigate, if CQL2
+should be extended through additional functions and operators to evolve into a backward-compatible expression language
+that is not restricted to boolean-valued expressions, but support expressions that result in other values, too, including
+geometries. Such expressions can be used, for example, in a style language or to define derived properties. In this case,
+a new requirements class for boolean-valued filter expressions would be added to the Standard.
--- a/cql2/standard/clause_1_scope.adoc
+++ b/cql2/standard/clause_1_scope.adoc
@@ -10,6 +10,6 @@ This document defines
 NOTE: The CQL2 grammar contains many language elements of a general purpose, spatio-temporal expression language.
 The focus of this version of the language are boolean-valued filter expressions. It is planned to investigate, if CQL2
 should be extended through additional functions and operators to evolve into a backward-compatible expression language
-that is not restricted to boolean-valued expressions, but support expressions that result in other values, too, including
+that is not restricted to boolean-valued expressions, but supports expressions that result in other values, too, including
 geometries. Such expressions can be used, for example, in a style language or to define derived properties. In this case,
 a new requirements class for boolean-valued filter expressions would be added to the Standard.
--- a/cql2/standard/clause_1_scope.adoc
+++ b/cql2/standard/clause_1_scope.adoc
@@ -7,9 +7,5 @@ This document defines
 * A text encoding for a CQL2 filter;
 * A JSON encoding for a CQL2 filter.
 
-NOTE: The CQL2 grammar contains many language elements of a general purpose, spatio-temporal expression language.
-The focus of this version of the language are boolean-valued filter expressions. It is planned to investigate, if CQL2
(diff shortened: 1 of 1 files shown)


cql2/standard/clause_1_scope.adoc
```diff
@@ -6,3 +6,10 @@ This document defines
 
 * A text encoding for a CQL2 filter;
 * A JSON encoding for a CQL2 filter.
+
+NOTE: The CQL2 grammar contains many language elements of a general purpose, spatio-temporal expression language.
+The focus of this version of the language are boolean-valued filter expressions. It is planned to investigate, if CQL2
+should be extended through additional functions and operators to evolve into a backward-compatible expression language
+that is not restricted to boolean-valued expressions, but support expressions that result in other values, too, including
```
pvretano, line 13: Should read ... " but supports expressions that result in other values, too, ...".  Otherwise the addition is good.

pvretano (approved):
Looks good.  Just one typo I noted.

jerstlouis:
Regarding:

>  extended through additional functions and operators to evolve into a backward-compatible expression language

Actually, the main change required is not to add any function and operators (the existing _Functons_ conformance class already supports custom functions), but to simply remove the current restrictions that prevent the use of:
- identifier expression (queryables)
- function calls
- numbers
- strings
- spatial literals
- temporal literals
- array literal
- arithmetic expressions

as the top-level rule that a CQL2 expression must conform to (currently `booleanExpression` in the BNF grammar).

cportele:
@jerstlouis - This is what "that is not restricted to boolean-valued expressions, but supports expressions that result in other values" states. To make good use of non-boolean-valued expressions in other contexts will almost certainly depend on additional standardized operators or functions beyond the boolean-valued operators. If you think the wording should be changed, could you propose a different wording and then we can decide in the next call? Thanks.

jerstlouis:
@cportele 

> To make good use of non-boolean-valued expressions in other contexts will almost certainly depend on additional standardized operators or functions beyond the boolean-valued operators

This is the part that I strongly disagree with.
Simply being able to use those other types of expressions (even if excluding function calls) would already make the language useful as _ParameterValue_ of _Symbolizers_.
For example, I could assign the `name` queryable as the text of a Label, or the `width` as the stroke width of a road -- all the sort of things that we have done in the Open Portrayal / Visual Style Editor tasks, and which _Symbology Encoding_ supports.

And whether operators / functions are standardized is really completely independent of the language definition. The mechanism to list functions, and ideally a mechanisms understand their arguments and semantics, should be sufficient -- I did not check yet whether arguments / semantics are included in the `/functions` end-point of Part 3. But at the very least all this can all be specified separately.

I would suggest the following tweaked text:

> NOTE: The CQL2 grammar contains all the necessary language elements of a general purpose expression language, including support for spatio-temporal values.
The focus of this version of the language are boolean-valued filter expressions. It is planned to potentially remove this restriction, while maintaining backward compatibility, allowing expressions that result in values of other data types, including geometries. Such expressions can be used, for example, in a styling language to specify parameter values, or to define derived properties. In this case, a particular use of CQL2 could specify the boolean-valued restrction separately, for example when used as a filter predicate expression, or as a styling rule selector. The usability of such an expression language, especially in the context of geometry types, would also greatly benefit from standardizing additional functions and/or operators e.g., to compute the geometry resulting from buffering or intersecting operations.

cportele:
Without standardizing the key operators like S_INTERSECTS a language like CQL2 would be much more difficult to use. Every client would have to inspect and understand the functions that an implementation provides first and everyone will use different names and arguments. For me, standardizing the key operators that are needed for the purposes that are in scope of the language is essential. It seems like we disagree on that. That said, your text looks fine, too. Let's resolve this in the next meeting.

jerstlouis:
@cportele Thanks. I agree that the divergence could be problematic, but with custom functions and domain-specific applications, with later extensions potentially clashing with existing custom functions, it is likely that at some point this may become inevitable. This is where a registry of functions with their semantics and URIs next to the function names in `/functions` could allow clients to make sense of the available functions even if divergence eventually occurs.

My main point of disagreement was that I believe there is already significant value in CQL2 as an expression language even excluding any functions / spatial operators, at least in the context of symbolizer parameter values and derived properties,
and that those functions / operators could still be standardized as a later part / extension slightly later when they are settled (though hopefully soon enough to avoid divergence).

cportele:
Meeting 2022-07-18: Everyone on the call is OK with both versions, Jerome has a preference for his, so we will use his version. @cportele will update and merge.

Issue #733, opened 2022-07-06
CQL2: Basic Spatial Operators issues


jerstlouis:
- In [7.5.2](http://docs.ogc.org/DRAFTS/21-065.html#_spatial_data_types_and_literal_values) where a link to the WKT standard is provided, it would be very insightful to specifically refer to **Section 7** (mention it in the link) of [_Simple Feature Access - Part 1: Common Architecture_](https://www.ogc.org/standards/sfa), which is where WKT is defined within that huge standard -- I wasn't clear at first that this was the right document where WKT is defined, since the title of the link differs significantly from the title of the standard. (at least 10,000 people are [wondering where WKT is defined](https://gis.stackexchange.com/questions/48269/where-can-i-find-the-wkt-standard-or-reference)).

- In [A.5.1 Conformance Test 22](http://docs.ogc.org/DRAFTS/21-065.html#_conformance_test_22), `ENVELOPE` is used with this syntax: `ENVELOPE(-180,-90,180,90)`. In Simple Features Access, _Envelope()_ is defined as a **method** to retrieve the envelop of the geometry. It is also a property of annotated text. It is _not_ defined for the WKT encoding. This [spatial4j issue](https://github.com/locationtech/spatial4j/issues/36) mentions the `ENVELOPE` syntax being introduced in Catalogue Service CQL this way: `ENVELOPE(` **`minX, maxX, maxY, minY`** `)` (with hatred for this order mentioned). Note that this is inconsistent with the current Abstract Test order. To avoid an enormous amount of confusion (already happening, as the current AT demonstrates), on one hand with the order inconsistent with BBOX used in _Features_, and on the other with envelope being a method rather than a geometry type, could we perhaps instead introduce a `BBOX()` that can take 4 or 6 parameters, consistent with _Features_?

- Because implementing support for _MultiPoints_, _Polygons_ (complete with holes), _MultiPolygons_, _LineString_, _MultiLineString_, _GeometryCollection_, and intersections for all these with all geometry types used on the server, from scratch still requires a great deal of effort, I would suggest to move these to the _Spatial Operators_ conformance class, and only keep `POINT` and the proposed `BBOX` (or `ENVELOPE` otherwise) along with `S_INTERSECTS()` in _Basic Spatial Operators_ conformance class.
   - If a comprehensive spatial library is used in the implementation, implementing all other operators alongside these more complex geometry types should not be difficult.
[...]

Issue #718, opened 2022-06-28
CQL2: Thoughts on arrays and `IN`


jerstlouis:
After completing an in-depth review of CQL2 (focused on `cql2-text` so far) as part of our implementation, the following thoughts on arrays and `IN` occurred to me:

As discussed in [this issue](https://github.com/opengeospatial/ogcapi-features/issues/705#issuecomment-1168126820), several of the conformance classes for extensions could actually have been defined separately by using functions as an extension mechanism.

One thing that stands out as not possible to define this way are the array literals which use the `[` and `]` syntax.

While I do love the `[` and `]` syntax for arrays (this is how we define arrays in JSON, eC, ECCSS/CMSS), I realized in CQL2(text) it feels somewhat out of place, because we actually have list of elements elsewhere in the language which use parentheses:
- Following the `IN` operator
   -  NOTE: It would also be useful to be able to use `IN` with an array on the right-hand expression
- In the WKT geometry definitions (points, linestrings, polygons, etc.)

I realize it's very late in the game, but I am throwing these suggestions for consideration just in case:
- Would it make sense to use `ARRAY(1, 2, 3)` instead of `[ 1, 2, 3 ]`?
   - Alternatively, should `IN` use the array syntax `IN [ 1, 2, 3 ]`?
- Should `IN` be usable with array types as a right-hand expressions, where you check containment of a single element inside a set?

I am also curious whether the original CQL had array and array operators, or if this is a recent addition?
If it is a recent addition, I imagine it is not widely implemented yet, so it might be worthwhile to make this change for greater consistency.

pvretano:
No.  The original CQL did not include arrays.  I think arrays were added as a result of issue #384.

jerstlouis:
@pvretano In that case, I would strongly suggest that:
- The syntax in cql2-text be `ARRAY(1, 2, 3)` to be consistent with everything else such as WKT geometry, time intervals, `IN`, and does not require the introduction of a completely new syntactical element (square brackets)
- `IN` accepts array expressions as a right-hand operand when both _Advanced Comparison_ and _Array Operators_ conformance classes are supported

pvretano:
Not opposed to it but we will need to discuss in the SWG.  As you pointed out, I was trying to go for an array syntax similar to JSON.

jerstlouis:
The `(1,2,3)` syntax is nice because it could be consistent with WKT and the `in` syntax.

[...]

Issue #705, opened 2022-06-14
Simplify the cql2-text grammar (future version improvements?)


jerstlouis:
This is feedback from trying to implement cql2-text.
Implementers (or at least us) face struggles with the current grammar.

I think it comes down mainly to these two things:
1. The extension conformance classes are defined as separate rules. I think it would be much easier to simply define new possible values for operators or pre-defined functions (using the same grammar rule as function calls) for operators using a function call syntax (i.e., array/spatial/temporal operators). This would cut down the number of rules drammatically, and I think would also allow to make the requirements in each conformance class clearer.
2. Some rules seeem to exist only to restrict the data types (e.g., numericExpression, characterExpression, temporalExpression...). However, this is purely a runtime concept, since the data type that a certain expression (e.g., a property) will evaluate to will depend on the queryables. Therefore I would not have used grammar rules (which are about the syntax) to make this distinction. Instead, I think what is needed for this is to have requirements and/or permissions that specify the interpretation if an unexpected data type is used in such a context.

I think simplifying these two aspects of the grammar would directly result in simpler parser implementations, greater ease of implementation and greater interoperability.

cportele:
Meeting 2022-06-20: It would be good to understand why this would result in an easier implementation. We need to discuss this in a meeting when @jerstlouis is present.

jerstlouis:
Thanks @cportele . I should be attending  the next meeting in a couple weeks.

As a summary, from a syntactic point of view, I think the two things I suggested above would result in fewer grammar rules (simpler grammar), and parser node classes would be a more direct / natural match to the rules. We would implement the function/operator name validation / data types checking separately from the parsing, since some of it is only known at runtime (e.g., available functions, queryable data types). e.g., in our implementation we have a _CQL2CallExp_ node class which we plan to use to handle the array / spatial / temporal operators which syntactically look like function calls. We are hand-writing an Recursive Descent parser, borrowing heavily from our [ECCSS/CMSS parser](https://github.com/ecere/ecere-sdk/tree/latest/compiler/eccss).

jerstlouis:
The following excerpt from our internal CQL2 design document mapping CQL2 conformance classes and providing a concise summary of the CQL2 syntax might be insightful. A simpler grammar could potentially closely match those `CQL2*` AST node classes to rules. We could eventually prototype such a simpler grammar together with railroad diagrams demonstrating the idea.

[...]

PR #745, opened 2022-07-18
Spatial operators

### LINKED ISSUE #734: CQL2: Spatial Operators Issues
In Table 2 of [7.6](http://docs.ogc.org/DRAFTS/21-065.html#spatial-operators), some of the mathematical definitions seem incorrect:
- `S_CONTAINS` is defined as `S_CONTAINS(a,b) ⬄ b CONTAINS a` -- First, I think it should have been the other way around, `a CONTAINS b`. But that definition would still be ambiguous. The SFA definition on page 39 (XXXIX) is defined as ` b.Within(a) `, because _Intersects_ and _Contains_ are convenience wrappers around _Disjoint_ and _Within_ respectively. _Within_ itself is unambiguously defined mathematically. Perhaps the intent was for this to be `S_CONTAINS(a,b) ⬄ b WITHIN a`. This is quite important because the actual meaning of _Contains_ is particularly counter-intuitive, as the [Wikipedia page](https://en.wikipedia.org/wiki/DE-9IM) states: 
   - The predicates _Contains_ and _Within_ have subtle aspects to their definition which are contrary to intuition. For example,[[10]](https://en.wikipedia.org/wiki/DE-9IM#cite_note-davis2007-10) a line L which is completely contained in the boundary of a polygon P is not considered to be contained in P. This quirk can be expressed as "Polygons do not contain their boundary". This issue is caused by the final clause of the Contains definition above: "at least one point of the interior of B lies in the interior of A". For this case, the predicate Covers has more intuitive semantics (see definition), avoiding boundary considerations.
- A related question is whether CQL2 should define _Covers_ and _CoveredBy_ as recommended instead of (or in addition to) _Contains_ and _Within_?
- `S_WITHIN` is defined as `S_WITHIN(a,b) ⬄ (a ∩ b = a) ∧ (I(a) ∩ E(b) ≠ ∅)`, however I believe that should be `S_WITHIN(a,b) ⬄ (a ∩ b = a) ∧ (I(a) ∩ E(b) = ∅)`, (`= ∅`, the interior of A should **not** intersect with the exterior of B for a to be within b), so their intersection is equal to the empty set, as it is on page 38 of [Simple Features Access](http://portal.opengeospatial.org/files/?artifact_id=25355).
- It would be useful to explain that `I(a)` means the interior of a and `E(a)` means the exterior of A (`B(a)` means the boundary of a, but that was only used in the DE9IM definitions).

cportele:
Closes #734 
Addresses two of three items in #733


cql2/standard/clause_8_encodings.adoc
```diff
@@ -32,7 +32,7 @@ The list of CQL2 keywords includes:
 * "BETWEEN"
 * "CASEI"
 * "DATE"
-* "ENVELOPE"
+* "BBOX"
```
jerstlouis, line 35: The keywords list is alphabetical so BBOX should be moved up two lines.

cportele: Thanks, fixed

cql2/standard/clause_7_enhanced.adoc
```diff
@@ -208,15 +208,33 @@ The spatial data types are (literal rule `spatialLiteral`):
...
 
-Existing representations are used for literal values:
+Existing representations are in general used for literal values:
 
-* Text: an https://portal.ogc.org/files/?artifact_id=25355[OGC Well-Known Text (WKT)] literal
-* JSON: a https://www.rfc-editor.org/rfc/rfc7946.html#section-3.1[GeoJSON geometry] object
+* Text: an OGC Well-Known Text (WKT) literal (see clause 7 of <<ogc06-103r4,Simple feature access - Part 1: Common architecture>>)
+* JSON: a GeoJSON geometry object (see clause 3.1 of <<GeoJSON,GeoJSON>>)
+
+The BBox data type is an exception as it is not specified in WKT. In the Text representation, the type is encoded as a `BBOX()` operator with four or six numerical arguments, depending on whether the coordinates include a vertical axis (height or depth):
```
jerstlouis, line 219: Should this perhaps say `BBOX()` **_literal_** rather than _operator_?

In CQL2 and WKT, `( )` are used everywhere so _grammatically_ speaking this is all the _function call_ syntax (that is the main point of #705), but conceptually there is a distinction between actual execution of a function / operator and a literal definition. e.g., in [CMSS](https://docs.ogc.org/per/18-025.html#_expressions), we use `( )` for function calls, `[ ]` for arrays  and `{ }` for type instantiation.

(related to #718 suggesting to use `( )` everywhere in CQL2 for array literals to be consistent with WKT and `IN` syntax, which really are already using a kind of array literals defined between `( )`).

cportele: Meeting 2022-08-01: @cportele to change to 'constructor' (and use this consistently with other cases like INTERVAL). Allowing properties in geometries is out-of-scope for this version. We might discuss this in a future version in the context of the future grammar simplification discussion.

cportele: Change applied in https://github.com/opengeospatial/ogcapi-features/pull/745/commits/d5527b9282c91a7322cfab135fcbc50a5f5ccbe2

cportele:
Meeting 2022-09-12: We will keep this open for reviews now. @pvretano has concerns about the use of "constructor", e.g., for CASEI. We will think about this topic until the next meeting.

jerstlouis:
@pvretano @cportele You may be interested in the conceptual work that we have been doing in _Styles & Symbology_ trying to UML-ify a [conceptual model for expressions](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/expressions.adoc) based on the Testbed 14 work, for potential inclusion in a revision of SymCore (or extension), intending it to fully cover CQL2 capabilities.

In this model,  the concept of **_InstanceExpression_** illustrates this idea of _constructor_.
In CQL2/WKT, this corresponds to the Geometry constructors and TimeStamp/TimeInterval usage.

**But for CASEI() and ACCENTI(), these rather correspond to [standard text manipulation functions](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/standardFunctions.adoc)** (called _caseInsensitize()_ and _accentInsensitize()_ there).

Also note the _InstanceLiteral_ inheriting from both _LiteralExpression_ and _InstanceExpression_, which is for instance expressions for which all values are fixed in the code (all member values are also _LiteralExpressions_).

Like WKT, CQL2 uses parentheses `( )` for function calls, constructors and lists/arrays (related to #718 where I suggest that should also be consistent for _ArrayExpressions_ used with the `IN` operator), whereas our CMSS and JSON uses curlies `{ }` for _InstanceExpressions_ and square brackets `[ ]` for _ArrayExpressions_.

Along with the [operators](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/operators.adoc), I believe these four UML diagrams ([expressions](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/expressions.adoc) / [operators](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/operators.adoc) / [standard functions](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/standardFunctions.adoc) / [geometry](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/UML/mermaid/geometry.adoc) -- currently missing _MultiPolygon_) fully covers the CQL2 functionality (and a bit more). From these, CQL2 does not have:
- the conditional expression and variable expression,
- the _(not)contains/startsWith/endsWith_ operators (it has `like` instead),
- the bitwise operators,
- the other text manipulation functions,
- the geometry manipulation functions, and
- the _s_covers()_ / _s_coveredBy()_ spatial relation functions.

cportele:
Meeting 2022-09-26: Change `CASEI` and `ACCENTI` not to be constructors, but "standardized functions". @cportele to update the PR including the terms and definitions.

@pvretano will have a look, if we can change all operators with a function-like syntax to functions in the grammar and define the functions separately.

PR #753, opened 2022-08-01
Issue-707

pvretano:
Update productions for characters and names to use productions from https://www.w3.org/TR/REC-xml/#charsets and https://www.w3.org/TR/REC-xml/#sec-common-syn.


cportele:
Meeting 2022-08-01: @pvretano will add text that explains which chars can start an identifier and which can be used in the rest of the identifier in plain English. After that he will merge the PR. @jerstlouis will review the details and create a new issue, if he identifies anything in this process.

PR #755, opened 2022-08-09
Updating mappings to specification elements of Part 1

ghobona:
This PR covers OGC API - Features - Part 1: Core v1.0.1 only.


ghobona:
@cportele Please review, edit and merge. 

Once the PR is merged, I will ask the OGC Infrastructure Office to enable the redirections on the server.

cportele (commented):
I think the docs-URIs should now use `docs.ogc.org`? And `https` instead of `http`?

There should also be an entry:

```csv
http://docs.opengeospatial.org/is/17-069r4/17-069r4.html,http://www.opengis.net/spec/ogcapi-features-1/1.0
```

There are more differences to 17-069r3.csv than expected, so before merging I want to write a script to check the entries.

cportele:
Replced by #881

PR #756, opened 2022-08-10
Updating mappings to specification elements of Part 2

ghobona:
Edited because there is no "fc-bbox-crs-valid-defaultValue" in the document.

There is however a "fc-bbox-crs-valid-default-value" in the document (which already has a mapping).


ghobona:
@cportele Please review, edit and merge.

Once the PR is merged, I will ask the OGC Infrastructure Office to update the redirections on the server.

cportele:
@ghobona - I just noticed one topic after merging: Should the docs-URIs be `http` or `https`? Probably `https`? If that is the case, I can update the file.

PR #774, opened 2022-10-24
Add the use of Last-Modified for optimistic locking.

pvretano:
I've added two conformance classes.  One for optimistic locking using ETags and one for optimistic locking using timestamps.  Still need to add words in the overview about using timestamps vs Etags and a discussion about strong and weak ETags as discussed in the issue.


cportele:
@pvretano 

- I can add the explanatory text to discuss the differences between the two approaches for feature data.
- We should probably merge #698 into this PR and then close that other PR?

cportele:
Meeting 2022-10-24:

* We will keep the two conformance classes as proposed.
* @cportele will update the text and merge #698 into this PR.

PR #775, opened 2022-11-21
Add support for the OPTIONS method.

pvretano:
Initial checking for adding support for the OPTIONS method.


extensions/transactions/create-replace-update-delete/standard/requirements/update/REQ_methods.adoc
```diff
@@ -0,0 +1,6 @@
+[[req_update_methods]]
+[width="90%",cols="2,6a"]
+|===
+^|*Requirement {counter:per-id}* |*/req/update/methods*
+^|A |A server SHALL declare support for the HTTP PATH method via the <<options,HTTP OPTIONS>> method.
```
cportele, line 5: PATH -> PATCH

extensions/transactions/create-replace-update-delete/standard/requirements/options/REQ_body.adoc
```diff
@@ -0,0 +1,6 @@
+[[rec_create-replace-delete_options_body]]
+[width="90%",cols="2,6a"]
+|===
+^|*Requirement {counter:req-id}* |*/req/create-replace-delete/options-body*
+^|A |The body of an OPTIONS request SHALL be empty.
```
cportele, line 5: I do not think this should be a requirement or even a recommendation. It is quite common the return the value of the Allow header in text/plain and I do not think this is a bad practice. Can we delete this and maybe add some informative text that makes clear that content in the response may or may not be provided?

extensions/transactions/create-replace-update-delete/standard/requirements/create-replace-delete/REQ_methods.adoc
```diff
@@ -0,0 +1,7 @@
+[[req_core_methods]]
+[width="90%",cols="2,6a"]
+|===
+^|*Requirement {counter:per-id}* |*/req/core/methods*
+^|A |A server SHALL implement one or more of the methods HTTP POST, PUT and/or DELETE for each mutable resource.
+^|B |A server SHALL declare which methods are supported for each mutable resource via the <<options,HTTP OPTIONS>> method.
```
cportele, line 6: Sub-requirement B makes the Options requirements class mandatory whenever Create/Replace/Delete is implemented. Which means that the requirements class is not necessary and all provisions can be moved to the Create/Replace/Delete requirements class. 

Either this, or remove B. However, in that case, the client needs to investigate the OpenAPI definition to find out if POST, PUT and/or DELETE is supported.

cportele (changes requested):
Looks good, except for two aspects/inconsistencies (and an editorial nit).

extensions/transactions/create-replace-update-delete/standard/requirements/create-replace-delete/REQ_methods.adoc
cportele: Meeting 2022-11-21: Move the OPTIONS text into Clause 6 and the Create, Replace, Delete requirements class.

extensions/transactions/create-replace-update-delete/standard/requirements/options/REQ_body.adoc
cportele: Meeting 2022-11-21: Remove the requirement and add informative text about the content (or not).

cportele:
Meeting 2022-11-21: @pvretano will merge the PR after the changes that we agreed on the review comments.

extensions/transactions/create-replace-update-delete/standard/requirements/options/REQ_body.adoc
pvretano: Added permission saying request may include a body but format is not defined in this specification and can be safely ignored.

extensions/transactions/create-replace-update-delete/standard/requirements/create-replace-delete/REQ_methods.adoc
pvretano: Removed OPTIONS requirements class and merged it into clause 6.

cportele (approved):
Looks good, I will merge.

PR #782, opened 2023-01-30
Refer to all spatial, temporal and array operators as standardized functions.

pvretano:
Resolve issue 776 by refering to all spatial, temporal and array "operators" as standardized functions.


cportele:
Meeting 2023-01-30: Merge this PR after the following change:

* `temporalLiteral` includes `intervalLiteral` which allows to include a property name or a function. This makes those rule non-literals and should be called using a different term, e.g. instance or object constructor. "literal" should only be used for constant values. @pvretano will review the use of "literal" throughout the spec and adapt the text as necessary.

We also discussed WKT literals like `POLYGON(...)`. Since CQL2 only allows WKT values (and no properties or arithemtic expressions) these are geometry literal instances only. A future extension could also add geometry constructor functions.

PR #784, opened 2023-01-30
Change array delimiters for parentheses rather than square brackets.

pvretano:


cportele:
Meeting 2023-01-30: @pvretano will merge after restoring the `function` rule with an explicit argument list. `arrayLiteral` needs to be renamed, see https://github.com/opengeospatial/ogcapi-features/pull/782#issuecomment-1408844870.

Spiritaine (approved):

PR #789, opened 2023-02-19
Fix "alpha" production in CQL2.

pvretano:
Fix "alpha" production in CQL2 to include missing characters and exclude the single quote.  I also removed some of the single-character productions like `colon = ":";` and simply used the literal string.  I think that simplified the grammar a bit and  makes it a little easier to read.

I also move `query.json` to the search extension.  I inadvertantly checked that in during some TB18 work.


pvretano:
@eseglem please review.  For some reason I was not able to add you to the list of reviewers,
@cportele also please revire and if you are OK with the changes merge.

cportele (approved):
Looks good to me. Waiting for feedback from @eseglem before merging.

eseglem (commented):
Definitely fixes the issue I was having with `alpha`. Swapped the update into the parser I have been working on and it is producing the expected results. 

I just had a few other comments brought on by the other updates. 

I may be missing something stating otherwise but I believe the two rules: `"'..'"` and `"'" ".." "'"` are functionally different. As one doesn't allow for white space and the other would. If that is not the case, may want to add some more clarification on that.

PR #790, opened 2023-02-19
Change names of some "literal" production because they are not stricly literals.

pvretano:
Change "instantLiteral" to "instantInstance", "temporalLiteral" to "temporalInstance", "spatialLiteral" to "spatialInstance" and "arrayLiteral" to "arrayInstance".


cportele (approved):
I have updated a few more places. Please check and merge, if my edits make sense.

cportele:
Meeting 2023-02-27: @pvretano will review and merge today.

jerstlouis:
The only thing here that is not consistent with the conceptual model for _Styles & Symbology_ is the `arrayInstance`.

_Arrays_ and _Instances_ are two fundamental types of expressions in it.
In programming languages, arrays are often implemented as some kind of container class, so it is not entirely distinct, but in JSON for example you use `[ ]` for arrays and `{ }` for instances.

**Could we simply call them `arrays` or `arrayExpressions` instead of `arrayInstances`, to avoid confusion?**

From [7.1.4.  Expressions](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#toc21):

download

download

For the _spatial instances_, they are essentially regular instances of a base _Geometry_ class in that model.
See [18.2.1.2 Geometry](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#toc59).

Somewhat related to all this, is recent consideration in extending CQL2 for the _Styles & Symbology_ encoding, and harmonizing WKT (which uses tupples without commas), INTERVAL (with start end separated by comma), the UML model that e.g., says that a POINT has a `coordinates` array, the need for explicitly setting properties and the css `<property>: value` syntax for instances. Also perhaps allowing `{ }` for instances and `[ ]` for arrays as an alternative there, for clarity.
See [21.1.6.  Expressions](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#toc70) for a lot of details about these considerations.

pvretano:
Just so I understand ... @jerstlouis you want me to rename `arrayInstance` to either `array` or `arrayExpression`.  Is that correct?  Also you want me to allow `{ }` for instances ... instances of what?  Also, you want me to allow arrays to be delimeted by either `( )` or `[ ]` ... is that correct?

I reviewed [21.1.6 Expression](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#toc70) and my initial comment is that people agreed to this?  It seem like you are allowing multiple ways to encode the same thing which I think is a bad idea.  Also, it seems you are trying to showhorn CQL2 into CSS which, again, seems a little awkward.

jerstlouis:
@pvretano 

> you want me to rename arrayInstance to either array or arrayExpression. Is that correct?

That is correct.

>  Also you want me to allow { } for instances ... instances of what? Also, you want me to allow arrays to be delimeted by either ( ) or [ ] ... is that correct?

Not in vanilla CQL2, at least not 1.0 (unless everyone think it's a good idea to make this change already?).
That is an open question for consideration right now in _Styles & Symbology_.

> instances of what?

With _Styles and Symbology_, as in #705, the idea is that from the grammar perspective, it doesn't matter instance of _what_.
Could be a Point, a Polygon, a Color, an Interval, an Instant, a user-defined instance from a complex schema. The syntax for an instantiation is consistent: `<class name> ( <member initializers> )` or `<class name> { <member initializers> }`.

> I reviewed [21.1.6 Expression](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#toc70) and my initial comment is that people agreed to this? It seem like you are allowing multiple ways to encode the same thing which I think is a bad idea.

That is a current topic of discussion in _Styles & Symbology_. We want to use and extend CQL2, but this is the logical conclusion of maintaining consistency with WKT, INTERVAL, and arbitrary instance expressions. Note that this is somewhat related to the _complex object_ support e.g., for Search/Queries.

> Also, it seems you are trying to shoehorn CQL2 into CSS which, again, seems a little awkward.

Note that this is _Cascading Cartographic Symbology Style Sheets_ and not _Web-CSS_.
We want to leverage CQL2, and we want the cascading aspect and general convenience of Web-CSS, and trying to figure out the best approach to extend CQL2 for that purpose in a way that feels natural.

I am accutely aware of the _multiple ways to encode the same thing_ which may seem like a bad idea, but so far this is the best we could come up with :) Ideas and suggestions welcome.

pvretano:
@cportele I changed `arrayInstance` to `array` as per the discusion above with @jerstlouis.  If no one objects at the 13-MAR-2023 SWG meeting then please go ahead and merge the PR.  Didn't want to unilaterally merge without people being aware of this -- albeit small -- change.

cportele:
@pvretano - I reviewed the changes, looks good to me.

Issue #783, opened 2023-01-30
Update examples ...


pvretano:
With all the recent (and upcoming) changes to CQL2 some of the examples are out of date and need updating.  I've created this issue and assigned it to me to do a sweep of the examples to make sure that they align with the current state of the text in the standard.

eseglem:
I came here to create an issue about the examples, so I hope this is all valuable input for this ticket instead. I have been working on a new parser and ran into several issues with the examples. 

- `example02.txt`: Uses `WILDCARD '#'` which is not in the current spec. 
- `example08.txt`: Has `'HH+VV+HV+VH'` which is probably intended as a character literal but `+` is `\x002B` and not in `alpha` as an acceptable character. This seems like it would actually be useful, and may be a spec issue? 
- `example20.txt` - `example22.txt`: Have dates in `"` rather than `'`

It would be very helpful if the text examples were directly parseable like the json is. They all contain explanations at the top which prevents reading them in as input without modifying them. A centralized readme with the explanations in a table would be a nice alternative. And several of the text files include alternatives within the file whereas the json is split apart into a and b files.

It would also be extremely useful for testing new implementations to have more official examples. There are many things that they don't cover and most are very simple. I think these would be nice to have: 
- Comparisons `<>` or `>=`
- `NOT BETWEEN`
- `NOT IN`
- `IS NULL` / `IS NOT NULL`
- Has 2 of 8 spatial operators
- Has 2 of 8 WKT 
- Has 3 of 15 temporal operators
- A timestamp instant. 
- All the array operators
- More complex identifiers with `_`, `.`, digits, and non ascii letters
- A `function` that isn't one of the defined operators
- Character literal with an escaped quote
- Boolean literal
- Arithmetic

If I can contribute anything useful here I am willing to help out as well. I am going to attempt to automatically generate some additional tests myself, but I am not sure how useful they may be to others. Their intent will be coverage on the grammar and not necessarily make any practical sense.

cportele:
Meeting 2023-02-15: @pvretano is working on a pull request and merge it once it is done. 

@eseglem - if you have examples, please send them to pvretano@cubewerx.com and he will include them, many thanks!

eseglem:
@cportele I have not generated much that is useful yet but I am still working on it. The first attempt was very naive and produced a lot of useless examples. Technically valid within the schemas but not actually usable. If / when I end up with anything useful I will share it.

cportele:
[...]

PR #794, opened 2023-02-25
Clean up minor inconsistencies in cql2 schemas.

eseglem:
While writing a parser I have spent a lot of time switching back and forth between the schema formats leading me to notice a few inconsistencies between the different formats so I have attempted to address them. I tried not to do anything dramatic, just align everything together a little better. 

The biggest issue I have had is `%` and `div` only existing in the cql2.bnf file, but there were lots of other little things that added up. I have attempted to provide detailed justifications for everything below. 

### Both cql2.json and cql2.yaml

- Added `%` and `div` to arithmetic op enums.
    - They are included in the bnf, as well as the text of the standard [clause_7_enhanced.adoc?plain=1#L790-L792](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/clause_7_enhanced.adoc?plain=1#L790-L792).

### cql2.json

- Switched `prefixItems` to `items` when it was used for objects of the same type. 
    - This better aligns with the yaml schema. 
    - It felt easier to read.
-  Replaced all `additionalItems` with `minItems` and `maxItems`.
    - `prefixItems` does not require all items in the array to exist. So, `minItems` is needed to require all of them.
    - More consistent within the schema and better aligns with the yaml schema.
    - The 2020 json-schema draft removed `additionalItems` so it seemed better to remove and be forward compatible. 
- Aligned function array items with yaml. `arrayLiteral` -> `arrayExpression`.
    - The rest of the items are `Expressions` in both the json and yaml so I figured this would be the one to adjust.
    - **Note:** Switching to `Literal` (or `Instance` per #790) types and adding `function` and `property` directly would better align to the bnf file and feels easier to read. But I did not want to do that without discussion. This would also apply to various other locations where similar things are done as well. 

### cql2.yaml

- Moved comments related to json `prefixItems` to `description` within the yaml.
    - Based on other examples I have seen this seems to be the more acceptable place for that information.
    - **Note:** A move to OpenAPI 3.1 would support `prefixItems` and no longer need these comments. I figured that would need additional discussion though. 
- Added `patternExpression` to `isLikeOperands`. 
    - While it is already a subset of `characterExpression` this aligns with the json better, and makes the description more useful. 
    - Would need to be included anyways for `prefixItems` if an OpenAPI 3.1 upgrade happened.
- Add `minItems` and `maxItems` to `inListOperands`. 
    - This aligns with the other schemas. I think it was just missing before. 
- Quoted all the arithmetic and binary operators for consistency within the schema. 
- Adjusted order of arithmetic and binary operator enums for consistency with json and bnf.

### DIFF
--- a/cql2/standard/schema/cql2.json
+++ b/cql2/standard/schema/cql2.json
@@ -101,14 +101,9 @@
       "type": "array",
       "minItems": 2,
       "maxItems": 2,
-      "prefixItems": [
-        {
-          "$ref": "#/$defs/scalarExpression"
-        },
-        {
-          "$ref": "#/$defs/scalarExpression"
-        }
-      ]
+      "items": {
+        "$ref": "#/$defs/scalarExpression"
+      }
     },
 
     "scalarExpression": {
@@ -156,16 +151,17 @@
       }
     },
     "isLikeOperands": {
-          "type": "array",
-          "prefixItems": [
-            {
-              "$ref": "#/$defs/characterExpression"
-            },
-            {
-              "$ref": "#/$defs/patternExpression"
-            }
-          ],
-          "additionalItems": false
+      "type": "array",
+      "minItems": 2,
+      "maxItems": 2,
+      "prefixItems": [
+        {
+          "$ref": "#/$defs/characterExpression"
+        },
+        {
+          "$ref": "#/$defs/patternExpression"
+        }
+      ]
     },
 
     "patternExpression": {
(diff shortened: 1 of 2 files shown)


cportele:
Meeting 2023-02-27: Thank you for the PR. The changes look correct. There was some issue with using `items` which is why it was changed to `prefixItems`, but we need to check what the issue was. @pvretano will review and validate the updated examples (see #783) and check, if everything is now consistent.

cportele:
Meeting 2023-04-10: @pvretano will review.

pvretano (approved):
The change from `prefixItems` to `items` is OK.

PR #802, opened 2023-03-13
Add additional content around escaping in string literals.

### LINKED ISSUE #717: CQL2 Escaping
Regarding escaping characters, currently CQL2 says:
- Single quotes `'` are repeated to escape them within string literals
- Backslash `\` is used to escape `_` and `%` in `LIKE` patterns

Comments / Questions:
- Repeating `'` vs. using `\` is somewhat inconsistent -- should `\'` also work for escaping single quote?
- Is it possible to support special escape sequence for characters such as new lines (`\n`) or tabs (`\t`) in string literals?
- Should it be possible to escape double quotes `"` within double-quoted identifiers (`\"`) following #707 changes?

pvretano:
- Add text indicating that escaping depending on the encoding being used
- Add text indicating that, depending on the back-end platform, additional processing of the string literal may be required.
- Add requirements for escaping in string literals in CQL2 text.

Closes #717

### DIFF
--- a/cql2/standard/requirements/cql2-text/REQ_escaping.adoc
+++ b/cql2/standard/requirements/cql2-text/REQ_escaping.adoc
@@ -2,9 +2,9 @@
 [width="90%",cols="2,6a"]
 |===
 ^|*Requirement {counter:req-id}* |*/req/cql2-text/escaping* 
-^|A |The escape character in a string literal SHALL be the backslash (`\\`).
-^|B |Embeded single quotations (`'`) in a string literal SHALL be escaped using a double single quotation (`''`) OR the backslash (`\\`) character.
-^|C | The server SHALL be able to parse the following escaped sequences for encoding control characters in a string literal:
+^|A |The escape character in a character literal SHALL be the backslash (`\\`).
+^|B |Embeded single quotations (`'`) in a character literal SHALL be escaped using a double single quotation (`''`) OR the backslash (`\\`) character.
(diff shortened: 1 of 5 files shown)


pvretano:
@cportele please review at the 13-MAR-2023 SWG meeting.  I made this a draft because I still need to add the content to the ATS.

cportele:
@pvretano - I reviewed the changes so far, they reflect what we have discussed in the issue.

cportele:
Meeting 2023-03-13: Changes look good so far. One aspect from the issue has not been addressed and maybe is part of another PR: the `whitespace` rule?

cql2/standard/requirements/cql2-text/REQ_escaping.adoc
eseglem, line 1: I think this could use a little more clarification. To me, 'string literal' is a bit ambiguous and could be referring to both `characterLiteral` and `identifier`. Specifically in A, is `\"` supposed to be allowed in the middle of an `identifier`? The discussion seems to indicate it should, but the current `identifierPart` does not allow for it. I think B and C may only apply to `characterLiteral` though.

eseglem (commented):
Sorry if I am jumping the gun a little bit with the comments. Been watching the repo to make sure I don't miss anything new or upcoming.

pvretano:
Thanks @eseglem. Not jumping the gun at all.  More eyes and brains = better standards! ;)

cportele (approved):

PR #803, opened 2023-03-26
Allow the use of PUT to create new resources

### LINKED ISSUE #595: Allow PUT to create resources?
This is revisiting #329. The topic came up again during the May 2021 sprint in the context of styles. See https://github.com/opengeospatial/ogcapi-code-sprint-2021-05/issues/36. 

At least for the general requirements class, a mechanism is needed to support a client proposal for an identifier, if the server wants to support it. PUT would be a natural mechanism for this.

The feature-specific requirements class could still restrict creating features to POST.

cportele:
Changes:

* This change adds the permission for server to support PUT for creating new resources.
* For features, a property `supportsNonAutogeneratedResourceIds: true` has to added to Collection resource.
* Clarification that to conform to the standard the server has to support POST, PUT or DELETE on some resource(s) endpoints for some users.
* Several inconsistencies were fixed (e.g. in requirement/permission URIs and anchors).

Closes #595.


extensions/transactions/create-replace-update-delete/standard/clause_6_create-replace-delete.adoc
pvretano, line 1: @cportele  I think this line needs a bit of clarification: 

"If the HTTP PUT method is executed on a non-existing resource in a collection, the server can add a new resource instance to a collection, if the collection supports that resource identifiers can be assigned by the client."

Perhaps we can word it like this:

"If the HTTP PUT method is executed on a non-existent resource in a collection, the server can add a new resource instance to that collection.  If the resource is already assigned an identifier and the server supports that resource identifiers can be assigned by the client, then the server can use the supplied identifier for the new resource instance.  Otherwise, the server can assign its own identifier to the new resource instance."

What do you think?

extensions/transactions/create-replace-update-delete/standard/requirements/create-replace-delete/REQ_methods.adoc
pvretano, line 1: Hmm .... I think this sentence needs a bit of work.  The original requirement was that a server needs to implement the oje or more of POST, PUT or DELETE at each resource endpoint.  You seem to want to change that to say that a server can implement POST on some resource endpoint and PUT or DELETE on some other resource endpoints.  Is that correct?  I'm not sure I see why we need to make this distinction especially since we already say that a client can use the OPTIONS method to see exactly which methods are implemented at a particualr resource endpoint.

cportele: The problem that I see is that POST, PUT or DELETE must be implemented "**for each resource**" (of the API). I think what we mean is that there must be some resources where POST, PUT or DELETE are implemented, it does not have to be on all resources. Where this is the case can be determined using OPTIONS. I agree, I also do not like the wording, it was an attempt to be more precise.

extensions/transactions/create-replace-update-delete/standard/clause_6_create-replace-delete.adoc
cportele: > Otherwise, the server can assign its own identifier to the new resource instance.

I think this would be in conflict with PUT. The server cannot assign a different URI and return a 201 on a PUT request.

cportele: Meeting 2023-04-23: Yes, the last sentence would conflict with PUT, so we can leave the text as it is.

extensions/transactions/create-replace-update-delete/standard/requirements/create-replace-delete/REQ_methods.adoc
cportele: Meeting 2023-04-23: Leave this open, we need to find clearer wording.

cportele: Meeting 2023-12-18: Change to "A server SHALL implement one or more of the methods HTTP POST, PUT and/or DELETE for each mutable resource."

PR #815, opened 2023-04-10
Allow extension/user-defined functions to be used in place of booleans.

### LINKED ISSUE #796: Functions as a boolean expression
It looks like CQL2 allows functions to be used in place of numeric, string, geometry, temporal, and array expressions.  I'm curious if it is intentional that booleans are excluded from this list.

I'm looking to support a number of custom filter functions that express things like "user has X permission."  This could be expressed as a CQL2 function like this:

```json
{
  "function": {
    "name": "permission",
    "args": ["download"]
  }
}
```

The current spec/schema doesn't consider a function a boolean expression, so one option is to include the function in a binary comparison expression:
```json
{
  "op": "=",
  "args": [
    {
      "function": {
        "name": "permission",
        "args": ["download"]
      }
    },
    true
  ]
}
```

We have a number of similar boolean function type filters that I'd like to be able to support in CQL2, and I would like to avoid having to evaluate this extra comparison expression (or to get rid of it in the parser).

pvretano:
Closes #796.

### DIFF
--- a/cql2/standard/schema/cql2.bnf
+++ b/cql2/standard/schema/cql2.bnf
@@ -174,7 +174,7 @@ arrayExpression = array
                 | propertyName
                 | function;
 
-# An array literal is a parentheses-delimited, comma-separated list of array
+# An array is a parentheses-delimited, comma-separated list of array
 # elements.
 array = "(" ")"
       | "(" arrayElement [ { "," arrayElement } ] ")";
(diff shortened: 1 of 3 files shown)


cportele (approved):
Looks good, including the other consistency changes in the BNF. I updated a comment to align with the changes.

PR #819, opened 2023-04-13
Clarify context of normative statements about link properties

### LINKED ISSUE #818: Required fields for links
I'm currently comparing STAC, STAC API, OGC API Features and OGC API Records link objects.
STAC API made the `type` recently a required field due to [req 13B](https://github.com/opengeospatial/ogcapi-features/blob/master/core/standard/requirements/core/REQ_fc-md-links.adoc) in OAFeat, but the schema doesn't list type as being required:
http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/link.yaml
Or does 13B only apply to self and alternate links as mentioned in req. 13A?

@pvretano In Records the schema only requires href, I think: https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/common/link.yaml 

In Commons the schema requires href and rel: https://github.com/opengeospatial/ogcapi-common/blob/master/19-072/openapi/schemas/link.yaml

I feel like there's a bit of alignment or clarification needed, not only in Features...?

cportele:
Closes #818

It is also redundant to require that `rel` is included in links which are identified by their `rel` - which only has potential to add to any confusion. Redundant reference to `rel` have been removed.

### DIFF
--- a/core/standard/requirements/core/REQ_fc-rel-type.adoc
+++ b/core/standard/requirements/core/REQ_fc-rel-type.adoc
@@ -2,5 +2,5 @@
 [width="90%",cols="2,6a"]
 |===
 ^|*Requirement {counter:req-id}* |*/req/core/fc-rel-type* 
-^|A |All links SHALL include the `rel` and `type` link parameters.
+^|A |All links in a `200`-response where `rel` is `self`, `alternate`, `next`, or `prev` SHALL include the `type` link parameter.
 |===
--- a/core/standard/recomendations/core/REC_fc-md-descriptions.adoc
+++ b/core/standard/recomendations/core/REC_fc-md-descriptions.adoc
@@ -2,8 +2,8 @@
 [width="90%",cols="2,6a"]
 |===
 ^|*Recommendation {counter:rec-id}* |*/rec/core/fc-md-descriptions*
-^|A |If external schemas or descriptions for the dataset exist that provide information about the structure or semantics of the data, a `200`-response SHOULD include links to each of those resources in the `links` property of the response (relation: `describedby`).
-^|B |The `type` link parameter SHOULD be provided for each link. This applies to resources that describe the whole dataset. 
+^|A |If external schemas or descriptions for the dataset exist that provide information about the structure or semantics of the data, a `200`-response SHOULD include links to each of those resources in the `links` property of the response (relation: `describedby`). This applies to resources that describe the whole dataset.
+^|B |The `type` link parameter SHOULD be provided for links in that response where `rel` is `describedby`.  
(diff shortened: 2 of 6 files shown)


core/standard/requirements/core/REQ_fc-rel-type.adoc
```diff
@@ -2,5 +2,5 @@
 [width="90%",cols="2,6a"]
 |===
 ^|*Requirement {counter:req-id}* |*/req/core/fc-rel-type* 
-^|A |All links SHALL include the `rel` and `type` link parameters.
+^|A |All links in a `200`-response where `rel` is `self`, `alternate`, `next`, or `prev` SHALL include the `type` link parameter.
```
m-mohr, line 5:
```suggestion
^|A |All links in a `200`-response where `rel` is `self`, `alternate`, `next`, or `prev` SHALL include the `type` link parameter.
```

What about `previous`? As far as I understood @pvretano, `prev` and `previous` are both allowed and act the same.

What about `last` and `first`, if used in OAFeat?

m-mohr (approved):
Thanks. Looks good, just one comment for consideration.

I did not check whether you found all potential occurences though.

core/standard/requirements/core/REQ_fc-rel-type.adoc
cportele: Yes, both `prev` and `previous` are registered by IANA - and basically synonyms. We only mention `prev`, since `prev` is what the current HTML standard uses. 

Apparently `previous` was from HTML 4. If you look into the current HTML 4 version, it also uses `prev`, so I would consider `previous` as "outdated".

`first` and `last` can of course be used, too. We do that in our implementation. However, that will typically require using an offset for paging and we had decided to leave the paging approach to the implementation and only require `next` links. See #23.

m-mohr: Thanks. This leaves me with a couple of questions, but we can also copy this to a new issue:

1. Should there be a recommendation to use prev oder previous? And that otherwise they are equivilent and all requirements (e.g. type is required) for prev also apply to previous?
2. Why is the type field only required for prev and next? I understand this requirement as follows: It doesn't say anything which links are required (just requiring next is fine). This req seems to just say that the type field is required for the mentioned rel types. If that's the case, I guess the type should also be required for all 5 pagination rel types, right? Or why do we need a type on prev and next, but not on the others?

cportele: Yes, we should discuss this in a separate issue. My thoughts:

I think you are talking about something like a paging extension where clients could expect `prev`, `first` and `last` links. In such an extension, I would not include `previous`, it is just a complication for clients.

Maybe we should not have included anything, even in examples, about `prev` in Core?

pvretano (approved):

Issue #736, opened 2022-07-12
Part 4: server supported functionnalities


benoitdm-oslandia:
From the spec, it is mentioned on http://docs.ogc.org/DRAFTS/20-002.html#feature-representations this condition "The server has schema constraints for features in a collection". 

Does it means the server could have no schema constraint? 

If this is the case, what should be the server HTTP response code? And, also, is there somewhere we have to describe the server capabilities (like the WMS GetCapabilites)?

cportele:
Meeting 2022-09-12: 

> Does it means the server could have no schema constraint?

We want to support servers that are "schema-less" (e.g., with a document data store are the backend), but many servers will have underlying schema constraints. Those servers should publish the schemas in order to help clients to create requests that are not rejected.

> If this is the case, what should be the server HTTP response code? 

If there is no schema constraint and the submitted resource meets the requirements of the media type (e.g., a valid GeoJSON feature), the resource will be created/updated with a 20x status.

> And, also, is there somewhere we have to describe the server capabilities (like the WMS GetCapabilites)?

The server has to publish its API definition, its Conformance Declaration and may publish the schemas, for example. Here is a link to the current proposal for publishing the schemas: https://github.com/opengeospatial/ogcapi-features/tree/master/proposals/schemas

cportele:
Meeting 2022-10-24: We should add an informative description about schemas (or the lack of them) in the document. @pvretano will work on a PR.

cportele:
Closed by #823

Issue #740, opened 2022-07-13
Schemas for Create/Replace/Update


cportele:
This is feedback from developing a client.

The spec currently states:

> The feature schema can and often will depend in the operation, so a Schema resource will publish the specific schema representations for the CREATE, REPLACE and UPDATE operations.

A typical case is that the server may require or prohibit the presence of an "id" member/property, depending on the id policy of the server and the operation. Or for update, all properties are optional.

However, the current text does not state, which difference from the schema of the returnables are allowed.

In our work, most APIs include some transformation between the data in the data store and the representation that is returned by the API. A simple example is a link to another feature in the same dataset. In the data store it is typically a foreign key column. In the API response we often include it as a link object with 'href' and 'title' properties. There are many other reasons for such transformations, e.g., that the feature representation should conform to some schema (e.g., INSPIRE) that differs somehow from the storage schema of the data.

The transformations will often not be bijective, so it is not an option to accept data in the returnables schema as payload of a create/replace/update request, because it cannot be mapped unambiguously to the database schema.

As a result, a client also needs to be able to request the feature data in a representation that directly matches the database schema in order to support editing feature attributes for a replace/update operation. Such a request is currently not specified and we had to implement it as a custom extension (we use a query parameter `schema=receivables` in the Feature request).

I think what we need to do in the standard is the following:

1. Add requirements about the differences between the returnables schema and the create/replace/update schemas and restrict those to the resource id and whether a property is required/optional. This supports clients that want to edit a feature so that it can be replaced/updated.

2. Decide whether we want to support other differences between the returnables schema and the create/replace/update schemas in an additional requirements class or if we consider this out-of-scope for Part 4 and whoever needs this (like us) has to implement their own extension. If added, the requirements class would have to support fetching a feature in the replace/update schema.

cportele:
Maybe we do not need separate schemas for Create/Update? JSON Schema (and the OpenAPI schema variant) specifies annotations `readOnly` and `writeOnly` that could be used:

> The boolean keywords readOnly and writeOnly are typically used in an API context. readOnly indicates that a value should not be modified. It could be used to indicate that a PUT request that changes a value would result in a 400 Bad Request response. writeOnly indicates that a value may be set, but will remain hidden. In could be used to indicate you can set a value with a PUT request, but it would not be included when retrieving that record with a GET request.

Spec reference: https://json-schema.org/draft/2020-12/json-schema-validation.html#name-readonly-and-writeonly

[...]

Issue #612, opened 2021-07-30
How to get the collection schema?


pvretano:
[Recommendation 2](https://docs.ogc.org/DRAFTS/20-002.html#feature-representations) of Part 4 specifies that the schema that should be retrieved from the `/collections/{collectionId}/schema` endpoint is the schema of a **feature**.  The example and requirements that follow [Recommendation 2](https://docs.ogc.org/DRAFTS/20-002.html#feature-representations) use a parameter named `type` to retrieve the `create`, `replace` or `update` schema of a feature.

My question is: How would someone get the schema of the collection so that it could be used to validate the response from `/collections/{collectionId}/items` endpoint?

In my implementation I have added the enumeration value `collection` for the `type` parameter so that `/collections/{collectionId}/schema?type=collection` returns the schema of the feature collection rather than of a feature. 
 However, this is just my implementation.  I created this issue so that we can discuss a mechanism for retrieving the collection schema.

cportele:
Here is how we have addressed this for now in ldproxy:

Create separate endpoints for the feature schema (`/collections/{collectionId}/schemas/feature`) and feature collection schema (`/collections/{collectionId}/schemas/collection`), because these are arguably different resources. That said, maybe it should be `.../item` rather than `.../feature`?

In the meeting on 2021-08-02 we decided to park the issue until we get to it in the discussion of part 4 or the new schema part.

cportele:
I think we can close this issue, it is no longer relevant. There was a decision that the schemas are an encoding-agnostic representation of the logical data schema.

Schemas for validation can still be linked to responses. A feature will link to the schema of a feature, a feature collection will link to the schema of the feature collection. These schemas can be hosted anywhere.

cportele:
Meeting 2023-06-19: This can be closed, it is no longer relevant as explained in the previous comment.

PR #823, opened 2023-04-24
Add informative clause about schemas.

pvretano:
Closes issue #736.


extensions/transactions/create-replace-update-delete/standard/clause_6_create-replace-delete.adoc
```diff
@@ -67,6 +67,15 @@ The API Description Document describes the HTTP status codes generated by that A
...
 
+==== Schemas
+
+This specification makes no assumptions about any schema constraints that a server may impose upon incoming features.  The intent is to support both "schema-less" servers (such as those back-ended by document data stores) as well as servers that have underlying schema constraints (such as those back-ended by an RDBMS).
+
+In the "schema-less" case, as long as the incoming feature meets the requirements of the declared media type (e.g. a valid GeoJSON feature) then the resource should be created/updated with a 20x status response.  
+
+In the case where the server imposes schema constraints, the server should publish schemas in order to help clients create requests that are not rejected.  How those schemas are published is out of scope for this standard although other parts of the OGC API Features suite of standard may address this topic.
```
cportele, line 76: > How those schemas are published is out of scope for this standard ...

This is not consistent with section `#feature-representations` which has requirements about schemas. But we could also merge this now as we will have to update all mentions of schemas as a result of #740 and also #612. We might also move all schema discussions directly to the [new Schemas extension](https://github.com/opengeospatial/ogcapi-features/tree/master/proposals/schemas), that this sentence alludes to.

cportele (approved):
Looks good, there is only one item for discussion (which we can also have later).

Issue #828, opened 2023-04-30
CQL2-JSON: `function` instead of `op` for casei/accenti, temporalPredicate, spatialPredicate, arrayPredicate?


jerstlouis:
In the JSON encoding for CQL2, would it make sense for the sake of consistency to still use `function` instead of `op` for those standard functions defined by these various requirements classes of the standard, just like the user-defined functions?

`casei()` and `accenti()` are essentially text manipulation functions returning a case/accent-insensitized version of the input string, and additional functions  such as `lowercase()` and `uppercase()` could be defined in the future.

In[ Styles & Symbology](https://opengeospatial.github.io/ogcna-auto-review/18-067r4.html#_4fbdadd9-8cd4-4d57-be49-a8e276e4b5ba) (for which we're defining a JSON encoding with a [JSON schema](https://github.com/opengeospatial/styles-and-symbology/blob/main/core/schemas/CartoSym-JSON.schema.json) based on CQL2-JSON), this would feel a lot more natural since we are not calling these operators at all, and these standardized functions are all listed right after and based on the generic concept of functions (which can include user-defined functions). In addition to spatial relation, temporal relation, array relation, we add additional text manipulation functions (_lowerCase, upperCase, concatenate, substitute, format_) as well as geometry manipulation functions (_intersection, union, difference, symDifference, buffer, convexHull, envelope_).

CQL2 already calls them functions in several places:

> The [Case-insensitive Comparison](http://docs.ogc.org/DRAFTS/21-065.html#rc_case-insensitive-comparison) requirements class adds a **standardized string function** to support case-insensitive string comparisons

> The [Basic Spatial Operators](http://docs.ogc.org/DRAFTS/21-065.html#rc_basic-spatial-operators) requirements class specifies minimal requirements for servers that support **standardized spatial comparison functions**.

> The [Temporal Operators](http://docs.ogc.org/DRAFTS/21-065.html#rc_temporal-operators) requirements class specifies requirements for servers that support standardized **temporal comparison function**. The list of **temporal comparison function** that must be supported is:

> The [Array Operators](http://docs.ogc.org/DRAFTS/21-065.html#rc_array-operators) requirements class specifies requirements for **standardized array comparison functions** for sets of values. The **array comparison functions** that must be supported are:

These requirements classes could also potentially be renamed to not talk about _operators_:

- "Basic Spatial Relation Functions"
- "Spatial Relation Functions"
- "Temporal Relation Functions"
[...]

PR #831, opened 2023-05-08
Fixes for #828

jerstlouis:


jerstlouis:
@pvretano @cportele

cportele:
Meeting 2023-05-22: @cportele and @pvretano to review, more reviews are welcome, too!

cportele (approved):
The PR implements the changes agreed in https://github.com/opengeospatial/ogcapi-features/issues/828#issuecomment-1538453680. It goes beyond that an in addition it also classifies the former spatial/temporal/array operators as spatial/temporal/array *relation* functions (the use of "relation" was not part of the previous decision), but I guess with a view to a future extension beyond filter expressions the use of "relation" makes sense.

cportele:
Meeting 2023-06-19: @pvretano will review today. If approved, we can merge the PR:

pvretano (changes requested):
The meaning of the term "relation" is very context dependent.  For example, in databases a "relation" exists between two database tables when one table has a foreign key that references the primary key of the other table.  Because of this I would say we should be careful how we use the term.
I would prefer that we just stick with "functions" ... "spatial functions", "array functions", "temporal functions", etc.

jerstlouis:
@pvretano Yes the relation is context dependent, but here a context is given: _spatial relation_ functions, _array relation_ functions.

If we just say "spatial functions", then I would expect to find a centroid or buffer calculation in there, so it seems too general.

"Relation" is the standard mathematics term to describe what all of these functions do:

https://en.wikipedia.org/wiki/Relation_(mathematics)

Yes, "relation" is also use in Databases with that other meaning:

https://en.wikipedia.org/wiki/Relation_(database)

But wouldn't you agree that here from the context, it is obvious which meaning is intended?

cportele:
Meeting 2023-07-03: It is just a terminology issue and has no impact on CQL2 as a language. Since there is no consensus on changing the previous decision we will stick to it and not use "relation function", just "function". We will make the change and then merge.

Issue #850, opened 2023-07-24
CQL2: (AT) Table 9 wrongly expects københavn matching kobenhavn


jerstlouis:
`københavn` NFD decomposition is still `københavn` (e.g., try here: https://dencode.com/en/string/unicode-normalization), so no non-spacing mark gets stripped and an `accenti()` comparison with `kobenhavn` should be _false_.

`ACCENTI` should NOT match it with `kobenhavn`, as it seems to be expected to several places in [Table 9](https://docs.ogc.org/DRAFTS/21-065.html#_conformance_test_23).

That would be consistent with the expected result in the last entry of Table 8:

`CASEI(name) IN (casei('Kiev'), casei('kobenhavn'), casei('Berlin'), casei('athens'), casei('foo'))` = `3`

where it only expects _Kiev_, _Beriln_ and _Athens_ to match.

Also as noted in https://github.com/opengeospatial/ogcapi-features/issues/847#issuecomment-1634591254 , [Requirement 9](https://docs.ogc.org/DRAFTS/21-065.html#accent-insensitive-comparison) is missing a clause about actually stripping the accents (nonspacing marks) after the normalization. Normalization by itself does not remove accents, it only ensures a particular way to encode them.

@cportele @pvretano @pcdion

jerstlouis:
I looked into this a bit more and found the following resources on **_character folding_**:

- http://unicode.org/L2/L2002/02186-foldings-0d6.html
- https://www.w3.org/wiki/I18N/CanonicalNormalizationIssues#Character_folding

Normalization followed by stripping some characters (e.g., non-spacing marks) will **not** take care of several of these types of character folding, such as replacing `ø` by `o`, or matching katakana and hiragana equivalent (e.g., `か` and `カ`), or matching `æ` to `ae`.

However, according to that report (first link), it seems that nothing has yet been standardized in Unicode to define character folding (beyond the decomposition mapping and stripping combining marks), and this is only at a technical report stage. There are 40 different types of character foldings enumerated in the rows of that table in _4.2 Specification of folding operations_, so the complexity of to implement all those cases of character folding is daunting. It states in the last column that _Data file specifying the mapping_ are TBD specifically for:
- Diacritic folding (includes stroke, hook, descender) (AccentFolding.txt **[TBD]**)
- Kana folding (KanaFolding.txt **[TBD]**)
- Letterforms folding (LetterFolding.txt **[TBD]**)

[...]

PR #857, opened 2023-09-11
Address issues 846, 847, and 850.

### LINKED ISSUE #846: CQL2: Unicode standard version now 15
[Requirements 8](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/requirements/case-insensitive-comparison/REQ_casei-function.adoc) and [9](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/requirements/accent-insensitive-comparison/REQ_accenti-function.adoc) explicitly reference Unicode version 14.0.

Should this be bumped to [version 15](https://unicode.org/versions/Unicode15.0.0/) released on September 13, 2022, and/or should this say "or a later version"?

### LINKED ISSUE #847: CQL2: Additional guidance on CASEI / ACCENTI
Having spent some amount of time digging through the Unicode resources to try to figure things out, I think it would be very useful to provide short and to-the-point additional informative guidance on what is meant instead of just referencing the standard for those who are implementing this manually. From my initial understanding:

**CASEI**

Essentially you want to use this file https://www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt (derived from UnicodeData.txt field 13 for simple lowercase mapping and https://www.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt for multiple characters mapping) and replace characters by the sequence on lines with a **'C'(ommon)** or a **'F'(full)**

**ACCENTI**

Essentially you need fields 3 and 5 from https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
and apply Normalization Form (K?) D (NF(K?)D: canonical (compatibility?) decomposition) by:
- Recursively replacing the code points by the **field 5 Decomposition Mappings** for those rows with canonical mappings (those mappings not prefixed by a `<`tag`>` from [Table 14](https://www.unicode.org/reports/tr44/#Character_Decomposition_Mappings), or any of them if applying NFKD (compatibility mappings); the decomposition type is also available in https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedDecompositionType.txt)
- Special rules apply to decompose Hangul syllables (which do not have a decomposition mapping set up).
There are some details in http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf [Section 3.12 of the Unicode Standard (from page 142)](http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf) and https://stackoverflow.com/questions/41309402/breaking-down-a-hangul-syllable-into-letters-jamo
The following C code implements the hangul syllable decomposition:
```C
if(codePoint >= 0xAC00 && codePoint < 0xD7B0)
{
   unsigned int syllable = codePoint - 0xAC00;
   unsigned int t = syllable % 28, v, l;
   syllable /= 28;
   v = syllable % 21, l = syllable / 21;
   add(0x1100 + l);
   add(0x1161 + v);
   if(t) add(0x11A7 + t);
}
```
- Applying the Canonical Ordering algorithm which is stable-sorting (e.g., bubble-sort) the decomposed mapping code points by the value of that combining class for any sub-string where the **Combining Class (field 3)** (also in https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedCombiningClass.txt) value is non-zero
(this step is necessary if there were combining marks in the source text; the Decomposition Mappings should otherwise already be in the correct order since version 2.1.9). -- NOTE: the canonical ordering will only matter for codepoints that do not get stripped, so it will not matter for any of the combining characters that are non-spacing marks)

then:
- Removing all _Nonspacing Marks_ (category `Mn`) (general category is **field 2** of UnicodeData.txt, also available in https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedGeneralCategory.txt). However, an exception should probably be made for some characters categorized as Mn, as some like the Japanese voicing marks (dakuten and handakuten) drastically change the consonants sound which would be analogous to changing k -> g or h -> p in latin alphabets (see [note below](https://github.com/opengeospatial/ogcapi-features/issues/847#issuecomment-1679819768)).

pvretano:
Closes #846
Closes #847


cportele (approved):
I added Unicode as a dependency to Basic CQL2, updated the ACCENTI tests to address also #850. As mentioned in that issue, I don't think that we should do more than we already have in this PR to the CQL2 standard. If there is a clear description somewhere that we can reference we should do that and add it to the bibliography (but I don't think we are aware of such a reference).

PR #867, opened 2023-11-01
Updates based on discussions at Oct 2023 Code Sprint

cportele:
List of changes:

- Change `/req/schemas/logical-model` into informative text (not testable).
- Add a "x-ogc-property-index" keyword in "Schemas" to express the sequence of the properties (in a coverage or for a feature encoding that is ordered, e.g. GML).
- Only reference JSON Schema 2020-12 (latest version).
- In `/rec/schemas/properties` clarify that no other keywords should be used.
- Add "x-ogc-unit" as a keyword in "Schemas" to declare the unit of measure of a numeric property.
- Add "x-ogc-unit" in examples.
- Additional ones, like the URI of a definition should be added in an extension, if needed.
- Do not use "id" as the property name of an "id" property in the examples.
- Improve the wording in some normative statements.
- Consider to add a "primary-interval" role with a proper interval representation (array of two instants).
- Fix typo (roadSegement -> roadSegment)
- Remove broken dependency in Clause 13.
- We should not restrict the data types of queryables, what is supported will depend on the filter language.  Add a note that filter languages may not support data types.

### DIFF
--- a/extensions/schemas/standard/clause_11_queryables.adoc
+++ b/extensions/schemas/standard/clause_11_queryables.adoc
@@ -40,12 +40,12 @@ In addition, a <<publisher-def,publisher>> may want to support <<queryable-def,q
 ^|*Requirement {counter:req-num}* |/req/{req-class}/{req}
 ^|A |A successful execution of the operation SHALL be reported as a response with a HTTP status code `200`.
 ^|B |For responses that use `application/schema+json` as the `Content-Type` of the response, the response SHALL conform to <<rc_schemas>>.
-^|C |No property SHALL be of type "object".
-^|D |The items of a property of type "array" SHALL either be strings or numbers.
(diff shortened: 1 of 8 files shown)


cportele:
Meeting 2023-11-06:

* Change "x-ogc-property-index" to "x-ogc-property-seq"
* For the languages to specify units of measurement, we also define a keyword "x-ogc-unit-lang" with the only value "UCUM" (the default). Add a note to clarify that communities or other standards can add more languages. 
* @cportele will also file an issue with the OGC-NA, the OGC unit register [http://www.opengis.net/def/uom](https://defs.opengis.net/vocprez/object?uri=http%3A//www.opengis.net/def/uom) seems to be incorrect (few values, but several entries exist multiple times).
* Once these changes have been addressed, @cportele can merge the PR.

PR #878, opened 2023-12-29
implement comment resolutions from ISO/DIS 19168-1 ballot

cportele:
See https://portal.ogc.org/files/?artifact_id=106797 (access restricted to members of the OGC Standards Working Group) for the list of comments and their resolutions.

### DIFF
--- a/core/standard/abstract_tests/ATS_class_core.adoc
+++ b/core/standard/abstract_tests/ATS_class_core.adoc
@@ -49,6 +49,8 @@ include::core/ATS_fc-md-items-links.adoc[]
 
 include::core/ATS_fc-md-extent.adoc[]
 
+include::core/ATS_fc-md-extent-multi.adoc[]
+
 ==== Feature Collection {root}/collections/{collectionId}
 
 include::core/ATS_sfc-md-op.adoc[]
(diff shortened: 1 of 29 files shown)


ghobona:
@cportele The phrase "OGC Web APIs" as used in the Commit https://github.com/opengeospatial/ogcapi-features/pull/878/commits/5b528b6f8631f82b4d17fd2d1e82062bc396e35f also implies that OGC owns the endpoints. So I suspect that the edits in the Commit do not address the concerns of the OAB.

cportele:
Why does "OGC Web API" imply that OGC "owns" the endpoints or the API?  Does "Web API" imply that the Web "owns" the API?

"OGC Web API" is used in all parts of the OGC API Features standard.  If I recall correctly, the term was originally introduced in the RFC comment resolution process of Common Part 1.  The OAB guidance does not propose an alternative term, so I do not see why it should be changed or to what.

ghobona:
@cportele I'm forwarding your question to @cnreediii for clarification of the OAB's request.

Cc: @gbuehler

cportele:
@ghobona @cnreediii @gbuehler - I suggest that we continue the discussion in https://github.com/opengeospatial/ogcapi-common/issues/336, not here, since this is a wider topic and not just related to this PR or OGC API Features in general. See my comment in that issue.

PR #879, opened 2024-01-01
Fix encoding query string parameters in examples

francbartoli:
Examples are inherited into pygeoapi openapi HTML page since the codebase references directly [https://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml](https://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml)

image

### DIFF
--- a/core/openapi/ogcapi-features-1.yaml
+++ b/core/openapi/ogcapi-features-1.yaml
@@ -644,11 +644,11 @@ components:
             $ref: '#/components/schemas/collections'
           example:
             links:
-              - href: 'http://data.example.org/collections.json'
+              - href: 'http://data.example.org/collections?f=json'
                 rel: self
                 type: application/json
                 title: this document
-              - href: 'http://data.example.org/collections.html'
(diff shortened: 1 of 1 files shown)


cportele:
@francbartoli - The current examples are valid examples, so why should this be changed? This seems to be a pygeoapi problem.

francbartoli:
@cportele they are misleading as compared to [this example 13.](https://docs.ogc.org/DRAFTS/17-069r5.html#_requirements_class_geojson)

cportele:
What is misleading? Both `?f=xyz` or `.xyz` are valid approaches - and other approaches are valid, too. See https://docs.ogc.org/DRAFTS/17-069r5.html#encodings.

cportele:
Meeting 2023-01-15: We close this issue, the text needs no change as discussed above. @francbartoli, if you think that anything still needs to be done, let us know.

Issue #882, opened 2024-02-01
CQL2: JSON isNull inconsistency


eseglem:
Going through and updating some of my stuff I noticed that `isNull` is very similar to other functions, but not quite the same. At least according to the spec files. 

In the [json](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/schema/cql2.json#L282) and [yaml](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/schema/cql2.yml#L192) it directly references the `isNullOperand`. 

```json
{
  "op": "isNull",
  "args": { "property": "name" }
}
```

For most (all?) of the other functions, `args` is an `Array` of the references. So it is a bit awkward for this one to be different. 

And interestingly the [example](https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/clause_6_basic_cql2.adoc?plain=1#L294-L299) in the docs actually shows it as an array. This makes me think the intent may already be for it to be an array. And the specs just need to be tweaked to match that intent. 

If so, I would be happy to submit a PR making the necessary adjustments.

pvretano:
@eseglem I think I made a boo-boo there!  It should be an array like all the other functions.  If you could create a PR, that would be much appreciated as I am jammed up with other stuff right now.

PR #884, opened 2024-02-02
CQL2: Example Updates / Fixes

eseglem:
## Background

I have been working on [pycql2](https://github.com/eseglem/pycql2) on and off for a while now. And I have been trying to utilize the examples in this repo for testing, but not all of them are currently usable for a few reasons:

- Examples 01 to 28 in the text files include extra descriptions / alternatives directly in the files. After that, they are all standalone. So it is also a bit inconsistent for them to be that way.
- There are some inconsistencies between the corresponding json / text files, so they do not query the same thing.
- A few had syntax issues.

I also noticed a bunch of nice examples in the `.adoc` files, which are not usable in an automated fashion since they are directly in the files. And there are actually a few issues in them as well. Pulling them out allows for syntax checking and for me to run some testing for parseability and consistency.

I have attempted to make them all usable and address various issues I have found in them. If there is too much going in this PR, I could split it in two. Figured it was worth starting the discussion as it is though. 

## Changes

- Removed "EXAMPLE" from 01-28 text files.
- Split any text files with "ALTERNATIVE EXPRESSION" to be 'a' and 'b' files.
    - Note: This is different from the -alt01 files, which are two ways of writing the exact same query. Vs two different queries that get the same results.
- For existing examples I matched up minor inconsistencies. At a high level:
    - In 01 json `scene_id` -> `landsat:scene_id` to match the text
    - In 04-06 made them all be `eo:` instead of some being `eq` or `ro`
    - in 08 text `geometry` -> `footprint` to match json, and some spaces
    - In 15 and 16 json `floor` -> `floors` to match the text
    - In 22 json switched it to an interval to match the text
    - In 24 json `location` -> `geometry` to match text
    - In 26-28,70,71,86 json fixed the `casei` and `accenti` syntax.
    - In 83 text fixed MultiPoint WKT
- Re-numbered examples in Clauses 6 and 7 to make things more consistent.
- Removed the full cql2 text and json examples from the `.adoc` files. And used `import` to include them in the same place.
    - The files are located in the `schema/examples/(json|text)/` folders at this point. If somewhere else makes more sense, can definitely change that.
    - There is still some text / json in the files. I believe anything I left behind is a partial example and not valid standalone.
- Fixed issues in the examples discovered after testing against them.
    - clause6 03 is part of why I noticed #882. I left it the array as is intended.
    - clause7 04,05 json `casei` and `accenti` signature changed
    - clause7 10,16 json `Polygon` needed additional brackets
    - clause7 12,13,17 text Datetime needed to be single quotes
    - clause7 13 was missing a comma in the json
    - clause7 15,18 text strings needed to be single quotes

## Additional Ideas

- If there is a desire to keep the text explanations for the original examples, perhaps they could be put in an annex?
- I was also thinking about trying to pull out the examples from the tables in the tests. That way they could also be check / directly used. But I think that is a bit more complex and deserves other discussion.


cportele (approved):
Thank you, @eseglem, for all the work. This is helpful for consistency and for future maintenance and reuse it would indeed also be good to make the ATS snippets also available as separate files, but we can also do this later.

PR #886, opened 2024-02-12
Add intermediate conformance class with the S_INTERSECTS operator.

### LINKED ISSUE #859: CQL2: conformance class for S_INTERSECTS for geometry, between Basic SO and SO classes
This comment is in the context of the Basic Spatial Operators being changed to only require support for S_INTERSECTS with bbox or Point, instead of also most other Geometry object. I think that's a good change to make the Basic Spatial Operators a simple to implement as possible, but it opens up a large gap between it and the Spatial Operators class. SO requires support for DE-9IM, which is much more difficult to implement, and in some cases, infeasible to implement if the backend database doesn't support those operators (as with Elasticsearch and OpenSearch). 

It would be nice if there were another spatial operators conformance class that has the same semantics as BSO perviously had, so that I can advertise that I support S_INTERSECTS with geometries, without having to "under advertise" I only support Basic, or be force to implement DE-9IM.

### LINKED ISSUE #885: CQL2: One remaining mention of spatial operators
https://docs.ogc.org/DRAFTS/21-065.html#cql2-text

In Requirement 22:

> The server SHALL be able to parse and evaluate all **spatial operators** encoded as a text string that validate against the BNF production fragments identified in the [Advanced Comparison Operators](https://docs.ogc.org/DRAFTS/21-065.html#rc_advanced-comparison-operators) requirements class.

This is the only remaining place that spatial operator is used, so probably _spatial functions_ should be used instead.

@pvretano  @cportele

pvretano:
Closes #859 
Closes #885


cql2/standard/annex_ats_extended-basic-spatial-functions.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+When:
+
+For each queryable `{queryable}` with a geometry data type, evaluate the following filter expressions
+
+* `S_INTERSECTS({queryable},BBOX(-180,-90,180,90))`
+* `S_INTERSECTS({queryable},POINT(7.02 49.92))`
+* `S_INTERSECTS({queryable},POINT(90 180))`
+* `S_INTERSECTS({queryable},BBOX(-180,-90,-90,90)) AND S_INTERSECTS({queryable},BBOX(90,-90,180,90))`
+
```
cportele, line 38: The tests should also cover all other data types that must be supported (MultiPoint, LineString, MultiLineString, Polygon, MultiPolygon, GeometryCollection), not just BBox and Point.

cportele (changes requested):

cportele (approved):
All issues have been resolved

Issue #544, opened 2021-04-01
Consider removing global list of queryables from Part 3


jampukka:
Global list of queryables is not useful before we have a resource for filterable cross-collection queries. I suggest we remove the resource `/queryables` (`/req/filter/get-queryables-op-global`) from Part 3 as well as Requirement 5 `/req/filter/filter-param-multiple-collections`.

There's nothing wrong with those requirements except they seem out of place. IMHO the extension that specifies cross-collection queries would be better place to introduce these requirements. There they could be conditional requirements to servers also implementing Part 3 the same way `filter-crs=https://` (`/req/filter/filter-crs-param`) only applies to servers also implementing Part 2.

cportele:
Meeting 2021-04-12: We need to take this into account in the discussion of the Schemas extension.

cportele:
Meeting 2021-06-07: The global list is not relevant for Part 3 as it is, because Part 3 only binds to the `items` endpoint. We need this for the Search/Query part or in the Tiles part that covers dataset tiles or any endpoint that works across multiple collections. We will move it to proposal folder of the Search/Query part.

cportele:
Closed by https://github.com/opengeospatial/ogcapi-features/commit/35e0c662076b34471918cb1933c1d3e52cce241e

ycespb:
>> The global [/queryables] list is not relevant for Part 3 as it is, because Part 3 only binds to the items endpoint. 

In which document/part  would there be a mechanism proposed to bind a queryable list to the /collections endpoint ?  I would assume that /queryables would be bound to the /collections endpoint (for collection search) and /collections/collection-id/queryables would be bound to /collections/collection-id/items.  This would not allow for a "global list" of queryables (for items) however.   Is there another way foreseen to allow discovering collection-search queryables and then "filtering of collections", and will this be in Part 3 "Filtering" ?   $5.3 of Part 3 defines the ".../queryables" relation as "referring to a resource that lists properties that can be used to query items in the collection represented by the link’s context".  Is there any reason to limit this "rel" to "items in the collection" and not allow this for "collections provided by the api" ?.

cportele:
[...]

PR #892, opened 2024-02-27
update definition of the "queryables" link relation

cportele:
See https://github.com/opengeospatial/ogcapi-features/issues/544#issuecomment-1363821650


cportele:
I now have doubts that this is a good idea. Items are always in a collection, so the existing wording seems to be correct. If we have a global queryables resource then the link's context will be a set of collections and each item will be in one of the collections. So, close this PR without merging?

PR #897, opened 2024-02-28
Add informative text about adding and removing properties from a resource.

### LINKED ISSUE #737: Part 4: PATCH with new property
From the spec, in the UPDATE section, the example http://docs.ogc.org/DRAFTS/20-002.html#_example_4 add a new property `primary_material` which does not seem to be in the original schema. 

IMHO, I think could be tricky to handle this case where we accept to add new properties to existing feature. 

By the way, does it mean all the feature of this collection will have this new property? And, in this case, what should be the default value?

pvretano:
Closes #737


cportele:
Code Sprint 2024-02-28: Merge

PR #904, opened 2024-03-10
CQL2 Text: update spatialLiteral rule

### LINKED ISSUE #800: CQL2 (Text): WKT Definition
I have noticed a few things which may be issues with the WKT definition in the grammar BNF. There are multiple definitions for WKT around, and I don't actually have access to the ISO spec, so it is hard to be certain what is correct and what isn't.

- `point` is defined with an optional Z coord: `xCoord yCoord [zCoord]` but none of the `*TaggedText` allow for a "Z" in the text. For example: `POINT Z (0 0 0)` would be the correct (or at least standard) way to write a 3D point in WKT. But would not be valid according to the grammar. 
    - Suggestion: Add `["Z"]` to each of the `*TaggedText` definitions.
    - This doesn't actually enforce "Z" matching up with 2D / 3D coordinates, but I was not worried about that part. At least shapely and PostGIS don't care if its there or not for input. But they will include it when outputting WKT.
- A `BBOX` does not seem to be a part of actual WKT but it is included in `spatialLiteral`. So, according to the grammar, you could have `GEOMETRYCOLLECTION (BBOX (0 0 0 0), BBOX (0 0 0 0 0 0))` but tools like shapely and PostGIS are not able to parse that.
    - Suggestion: Create a `geometryLiteral` without `bboxTaggedText` and use it for `geometryCollectionText`
- The grammar does not support `EMPTY` as an option. So `POLYGON EMPTY` would not be parseable. This may be intentional, but I could see some value in having it: `S_EQUALS("my_geom", POLYGON EMPTY)`. And if nothing else, it would seem to be a compatibility issue. 
    - Suggestion: Add `"EMPTY"` to each of the `*TaggedText` definitions.
    - If it were added to `*Text` that would allow weird things like `POLYGON (EMPTY, EMPTY)` which also doesn't work. 

I don't know if the following are actually incorrect or not but they do seem to lead to compatibility issues. It doesn't appear to be possible to use some values which are valid in the grammar as input to many tools. It is also an issue when trying to go back and forth between `cql2-text` and `cql2-json`. 

- Since `{"," point}` is repeated 0 or more times, the grammar allows for `LINESTRING(0 0)`. But that isn't a valid LineString in GeoJSON and tools like shapely and PostGIS throw an error when trying to parse it. They require at least 2 points. 
    - Suggestion: Use `lineStringText = "(" point "," point {"," point} ")";` Requiring at least 2 points..
- The grammar also allows for polygons with 1-3 points. Like `POLYGON((0 0, 1 1))` which also cannot be represented by GeoJSON or parsed by other tools. They require at least 4 points. (And the last point needs to be the same as the first, but I don't think that part can be represented in BNF.)
    - Suggestion: Add `linearRingText = "(" point "," point "," point "," point {"," point } ")";` to require at least 4 points and use it instead of line `polygonText =  "(" linearRingText {"," linearRingText } ")";`
- The GeoJSON spec states "implementations SHOULD avoid nested GeometryCollections" so things end up in a weird spot if the WKT allows for it, but the GeoJSON implementation does not.
    - Personally, I would prefer to see it not allow nesting, but it also seems to be valid WKT. So I may be in the wrong here. 
    - I also just noticed, while writing this up, that the cql2.json spec doesn't include GeometryCollection at all. I had mostly ignored that part of the spec and have been going directly from the GeoJSON RFC and https://github.com/geojson/schema.

All of that combined would be something like this:

```
spatialLiteral = geometryLiteral
               | geometryCollectionTaggedText
               | bboxTaggedText;

geometryLiteral = pointTaggedText
                | linestringTaggedText
                | polygonTaggedText
                | multipointTaggedText
                | multilinestringTaggedText
                | multipolygonTaggedText;

emptySet = "EMPTY"

pointTaggedText = "POINT" ["Z"] ( pointText | emptySet  );
linestringTaggedText = "LINESTRING" ["Z"] ( lineStringText | emptySet  );
polygonTaggedText = "POLYGON" ["Z"] ( polygonText | emptySet );
multipointTaggedText = "MULTIPOINT" ["Z"] ( multiPointText | emptySet );
multilinestringTaggedText = "MULTILINESTRING" ["Z"] ( multiLineStringText | emptySet );
multipolygonTaggedText = "MULTIPOLYGON" ["Z"] ( multiPolygonText | emptySet );
geometryCollectionTaggedText = "GEOMETRYCOLLECTION" ["Z"] ( geometryCollectionText | emptySet );

pointText = "(" point ")";
lineStringText = "(" point "," point {"," point} ")";
linearRingText = "(" point "," point "," point "," point {"," point } ")";
polygonText = "(" linearRingText {"," linearRingText} ")";
multiPointText = "(" pointText {"," pointText} ")";
multiLineStringText = "(" lineStringText {"," lineStringText} ")";
multiPolygonText = "(" polygonText {"," polygonText} ")";
geometryCollectionText = "(" geometryLiteral {"," geometryLiteral} ")";
```

cportele:
Closes #800. 

As discussed in https://github.com/opengeospatial/ogcapi-features/issues/800#issuecomment-1987256292, all suggestions are included except the proposal to support EMPTY in WKT geometries.

In addition, the following changes are included:

- change version and document generation to 21-065r1
- add missing requirements for the `basic-spatial-functions-plus` requirements class in the encoding requirements classes
- add missing conditional dependencies in the encoding ATSs
- fix a spelling mistake in the ATS
- fix an incorrect requirement identifier

### DIFF
--- a/asciidoctor.json
+++ b/asciidoctor.json
@@ -74,7 +74,7 @@
                 ]
         },
         "5":  {
-                "OGC_doc_number": "21-065",
+                "OGC_doc_number": "21-065r1",
                 "urn":"http://www.opengis.net/doc/IS/cql2/1.0",
                 "title":"Common Query Language (CQL2)",
                 "path":"cql2/standard/21-065.adoc",
--- a/cql2/standard/21-065.adoc
+++ b/cql2/standard/21-065.adoc
@@ -33,8 +33,8 @@
 |Approval Date:  <yyyy-mm-dd>
 |Publication Date:  <yyyy-mm-dd>
 |External identifier of this OGC(R) document: http://www.opengis.net/doc/IS/{standard}/{m_n}
-|Internal reference number of this OGC(R) document:    21-065
-|Version: {m_n}.0-rc.1
+|Internal reference number of this OGC(R) document:    21-065r1
+|Version: {m_n}.0-rc.2
 |Latest Published Draft: n/a
 |Category: OGC(R) Implementation Specification
 |Editors: Panagiotis (Peter) A. Vretanos, Clemens Portele
--- a/cql2/standard/annex_ats_basic-spatial-functions-plus.adoc
+++ b/cql2/standard/annex_ats_basic-spatial-functions-plus.adoc
@@ -20,7 +20,7 @@ The term "geometry data type" is used for the following data types: Point, Multi
 |===
 |Test id: | /conf/{conf-class}/{conf-test}
 |Requirements: | <<req_{conf-class}_spatial-predicate,/req/{conf-class}/spatial-predicate>>, <<req_{conf-class}_spatial-functions,/req/{conf-class}/spatial-functions>>, <<req_{conf-class}_spatial-functions,/req/{conf-class}/spatial-data-types>>
-|Test purpose: | Test the S_INTERSECTS spatial comparison function with points, multi-points, line strings, multi-line string, polyons, mult-polygons, geometry collections and bounding boxes.
+|Test purpose: | Test the S_INTERSECTS spatial comparison function with points, multi-points, line strings, multi-line string, polyons, multi-polygons, geometry collections and bounding boxes.
 |Test method: | 
 Given:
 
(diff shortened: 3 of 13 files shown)


pvretano (approved):
@cportele looks good to me.

PR #929, opened 2024-06-17
Initial check in of OGC API - Features - Part 8: Sorting

pvretano:


cportele:
Meeting 2024-07-01: @pvretano will add a note about Sortables (see previous comment) and then merge.

PR #933, opened 2024-06-25
fix broken links

ghobona:


cportele (approved):
Thanks @ghobona for catching and fixing these. I checked the updated links and they are all correct.

PR #935, opened 2024-06-29
prepare CQL2 for publication

cportele:


pvretano (commented):
@cportele ... look good to me.

PR #936, opened 2024-07-02
Final OGC staff review of CQL2

ghobona:


cportele (approved):
Thanks @ghobona, all fixes look good. 

I have also escaped the hashes (I do not fully understand why some backslashes are required and some not, but it now renders correctly).

Apologies for misspelling your name.

ghobona:
@cportele No worries about the mispelling.

ghobona:
@cportele Thanks for approving the PR.

PR #937, opened 2024-07-08
Part 3: final review for publication by ogc staff

ghobona:


ghobona:
@cportele Please note that one observation I was not sure about is that Conformance Test 3 says:
assert that the value of the $schema member is "https://json-schema.org/draft/2019-09/schema" or "https://json-schema.org/draft/2020-12/schema" 

The associated Requirement 3 says: For responses that use application/schema+json as the Content-Type of the response, the response SHALL have the following characteristics: The property $schema is https://json-schema.org/draft/2020-12/schema.

cportele (commented):
Looks good to me

cportele:
We now only support JSON Schema 2020-12, I have updated the test.

PR #952, opened 2024-09-23
Change arrayExpression->arrayOperand and array->arrayExpression.

### LINKED ISSUE #951: CQL2-JSON: "arrayExpression" was confusing
I believe that `minItems: 2` and `maxItems: 2` here is a left-over bug:

https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/schema/cql2.json#L258

pvretano:
Closes #951

### DIFF
--- a/cql2/standard/schema/cql2.bnf
+++ b/cql2/standard/schema/cql2.bnf
@@ -11,6 +11,8 @@
 #              portele[at]interactive-instruments.de  
 # 02-JUL-2024  pvretano[at]cubewerx.com               1.0.0 (final release)
 #              portele[at]interactive-instruments.de  
+# xx-xxx-2024  pvretano[at]cubewerx.com               1.0.1-SNAPHOST
+#              portele[at]interactive-instruments.de  
 
 #=============================================================================#
 # A CQL2 filter is a logically connected expression of one or more predicates.
--- a/cql2/standard/schema/cql2.bnf
+++ b/cql2/standard/schema/cql2.bnf
@@ -178,16 +178,16 @@ temporalFunction = "T_AFTER"
 # a Boolean value of true.  Otherwise the function returns false.
 #=============================================================================#
 arrayPredicate = arrayFunction
-                 "(" arrayExpression "," arrayExpression ")";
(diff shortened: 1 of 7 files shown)


cql2/standard/schema/cql2.json
```diff
@@ -250,22 +250,22 @@
           "type": "string",
           "enum": ["a_containedBy", "a_contains", "a_equals", "a_overlaps"]
         },
-        "args": {"$ref": "#/$defs/arrayExpression"}
+        "args": {"$ref": "#/$defs/arrayOperand"}
       }
     },
-    "arrayExpression": {
+    "arrayOperand": {
```
jerstlouis, line 256: @pvretano I think this should be `arrayOperands` plural to be consistent with the other rules and to highlight the fact that this represents two operands for the arra predicate functions.

cql2/standard/schema/cql2.bnf
```diff
@@ -178,16 +178,16 @@ temporalFunction = "T_AFTER"
 # a Boolean value of true.  Otherwise the function returns false.
 #=============================================================================#
 arrayPredicate = arrayFunction
-                 "(" arrayExpression "," arrayExpression ")";
+                 "(" arrayOperand "," arrayOperand ")";
```
jerstlouis, line 181: @pvretano Now I'm confused... Maybe this is is what was contributing to my earlier confusion.

Why does this take 2 `arrayOperand`?

cql2/standard/schema/cql2.json
cportele: I agree, this should be plural. This applies also to cql2.yaml.

cportele (changes requested):
In addition to changing "arrayOperand" in JSON/YAML to "arrayOperands", we also need to update the version information:

- In cql2.bnf we should add an entry in the history list at the top (which might be a placeholder until we tag a version for a 1.0.1 release).
- In cql2.yaml version should be "1.0.1-SNAPSHOT" (until we publish a 1.0.1).

cportele:
Meeting 2024-10-07: @cportele will make the changes and then merge the PR.

PR #960, opened 2024-10-21
provide schemas for both OpenAPI v3.0 and v3.1

cportele:
This is a follow-up to #948. For convenience, maintain schemas for both OpenAPI v3.0 and v3.1. Only Part 1 examples will use the v3.1 schema for now, others still use OpenAPI v3.0.

### DIFF
--- a/core/examples/README.md
+++ b/core/examples/README.md
@@ -3,6 +3,6 @@
 The sub-folders contain examples of responses of implementations of 
 OGC API - Features - Part 1: Core.
 
-* [OpenAPI examples](openapi)
+* [OpenAPI 3.1 examples](openapi)
 * [JSON/GeoJSON examples](json)
 * [XML/GML examples](xml)
--- a/core/examples/openapi/README.md
+++ b/core/examples/openapi/README.md
@@ -2,13 +2,15 @@
 
 This folder includes two complete examples of an OpenAPI definition for a Web API implementing OGC API Features (the Core, HTML, GeoJSON and OpenAPI 3.1 conformance classes).
 
-The [first example (ogcapi-features-1-example1.yaml)](ogcapi-features-1-example1.yaml) is a generic example that uses path parameters to describe all feature collections and all features. This OpenAPI definition does not provide any details on the collections or the feature content. This information is only available from the feature collection metadata. The API is also on [SwaggerHub](https://app.swaggerhub.com/apis/cportele/ogcapi-features-1-example1/1.0.0).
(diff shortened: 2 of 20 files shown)


cportele:
Meeting 2024-10-21: @cportele to explain the reasons and the structure with the consolidated OpenAPI 3.0 file and then merge.

Sam-Bolling:
> For convenience, maintain schemas for both OpenAPI v3.0 and v3.1

Is there a summary of context about this decision?  The OGC API - Connected Systems (CSAPI) Part 1 and Part 2 standards are in a precarious timeframe where it appears to make technical value to release based on OAS 3.1, but as-is, that would break the declaration of OGC CSAPI as an extension of the OGC API - Features standard.  Are there recommendations from the OGC Features SWG to SWGs with an OGC API standard designed to be an extension of OGC API - Features that is currently in its public review period?  It appears the draft OGC CSAPI was originally designed based on OAS 3.1, but has been influenced by the OGC process to "down-compile" to OAS 3.0.  If the OGC Features SWG has a new message and way forward on the topic today, the timing feels close but right if there is a chance that extension standards like OGC CSAPI now have an opportunity to focus more on OAS 3.1

Respectfully,

Sam

cportele:
@Sam-Bolling ,

there are two aspects of the use of OpenAPI in OGC API Standards (I think in general, but for sure in Features):

1. OpenAPI fragments are used to normatively specify characteristics of the API building blocks (content schemas, query or path parameters, etc.). In Part 1 v1.0 OpenAPI 3.0 was used (v3.0.3). In Part 1 the plan is to change to OpenAPI 3.1 fragments for this purpose.

2. OpenAPI definitions can be used by API instances to document the API definition. There will be separate requirements classes for OpenAPI 3.0 and OpenAPI 3.1. To support reuse of the API building blocks the idea is to publish the OpenAPI components of the API building blocks for both versions of OpenAPI. The differences are minimal and as far as we can see right now restricted to the use of "null" as a type in OpenAPI 3.1 vs the use of "nullable" as an attribute of a property in OpenAPI 3.0.

In my view, Connected Systems is free to use only OpenAPI 3.1. The use of OpenAPI 3.0 in Features Part 1 v1.0 does not put any obligation to use a specific OpenAPI version on any requirements class that has a dependency on Features. The only impact is that any Connected Systems API that publishes its API definition in OpenAPI 3.1 has to use the type "null", not the attribute "nullable" in the schemas of the API building blocks from Features. That is possible already with v1.0 of Features Part 1. The API just doesn't conform to the OpenAPI 3.0 conformance class, but that is perfectly fine and why specific API definition languages/versions are in separate requirements classes.

PR #986, opened 2025-05-04
Schemas: use format for more specific data types

### LINKED ISSUE #946: Part 5: The limitations of JSON data types
The document says:

> To use a schema for data validation, the schema must be converted into a schema representation suitable for validating data in the specific data format. For example, an XML Schema that is a GML application schema or a JSON Schema for GeoJSON or the draft OGC Features and Geometries JSON (JSON-FG).
> [...]
> The main reasons for using JSON Schema are:
> [...]
> - JSON data types (string, number, boolean, array, object, null) are simple and easy to understand; 

I'm working on something very similar in a project called fiboa, see https://github.com/fiboa/schema
My observation from this work, mapping JSON Schema to e.g. GeoParquet, is that it's not that easy as it sounds.
Especially the number data type is diffiult. 

- It's not clear how to map number to more granular data types, e.g. uint8, int8, uint64, int64, etc. As JSON has no limitations with regards to the number size, you may always need to choose the biggest data type possible. That's not ideal.
- There are strongly typed data/languages. Something like `type: ["number", "string", "array"]` may not be easy or impossible to represent
- The number type in JSON Schema doesn't include NaN and +/-infinity. How is this meant to map to other languages?

cportele:
closes #946


cportele:
Meeting 2025-05-05: merge after resolving conflicts

PR #988, opened 2025-05-05
Schemas: Add table of "x-ogc" elements.

### LINKED ISSUE #945: Part 5: Add table of x-ogc- properties
I think it would be valuable to add a table of all the pre-defined additions to JSON Schema, i.e. with the `x-ogc-` prefix.
It would help to understand schemas more easily, especially as some schemas consist of more proprietary properties than JSON Schema vocabulary.

PS: Generally, it feels like the number of additions to JSON Schema is a bit excessive.

pvretano:
Closes #945


cportele (approved):
LGTM

PR #1001, opened 2025-06-02
Schemas: QUDT in bibliography, grammar fixes, clarifications and fixed bad spaces

jerstlouis:


cportele (approved):
LGTM, except for the general QUDT reference in the bibliography. The only QUDT source that is used is already a normative reference in clause 3. 

I do not see a value in the addition to the bibliography. If we include the QUDT website in the bibliography we probably have to reference a larger number of other websites, too.

jerstlouis:
@cportele Regarding the bibliography, this is because the normative reference is specifically to the Unit vocabulary, but QUDT can also be used for the `x-ogc-definition`. We had discussed this before where we had changed the normative reference to the more general QUDT ontology, and it might have been your suggestion to keep the normative reference to the units only, but add the general ontology to the bibliography instead (since the mention of QUDT for `x-ogc-definition` does not appear in the requirements text).

cportele:
@jerstlouis - But there can many many other sources for definitions including FACC, DFDD, INSPIRE register, and many others. We clearly do not want to list them all or imply a preferred source.

cportele:
Meeting 2025-06-02: @jerstlouis will remove the QUDT reference from the bibliography, then we will merge the PR.

PR #1002, opened 2025-06-02
8-core-roles: renaming to Property roles

jerstlouis:
- More generic as it could apply to access mechanisms other than OGC API - Features

### DIFF
--- a/extensions/schemas/standard/clause_8_core_roles_for_features.adoc
+++ b/extensions/schemas/standard/clause_8_core_roles_for_features.adoc
@@ -1,8 +1,8 @@
-:req-class: core-roles-features
+:req-class: property-roles
 [#rc_{req-class}]
-== Requirements Class "Core roles for features"
+== Requirements Class "Property roles"
 
-The Requirements Class "Core roles for features" specifies additional property roles for feature data.
+The Requirements Class "Property roles" specifies additional property roles for data collections.
 
 [cols="2,7",width="90%"]
 |===


jerstlouis:
@cportele The primary use case I was seeing here is for declaring the `geometry` role for collections accessible using other OGC APIs and supporting a CQL2 filter in particular.

For example, `/coverage?filter=S_INTERSECTS(geom,POLYGON((-64.8 32.3,-65.5 18.3,-80.3 25.2,-64.8 32.3)))` (or also for `/map`).

In this context, the geometry gets evaluated on a per pixel / per cell basis while retrieving or rendering the data.

Even though the _examples_ are features-specific, I thought we had addressed most of the language with #1003 so that you can read this with this OGC API - Maps or OGC API - Coverages use case in mind, and it worked well. That was our intent anyways when reviewing this / proposing these changes.

cportele:
@jerstlouis - But geometry properties are declared through the "format" keyword, not a role. So for the CQL2 filter, the roles are not relevant.

The wording in that clause is feature-specific, not just the examples.

jerstlouis:
@cportele But isn't `"x-ogc-role" : "primary-geometry"` relevant to know which is the primary geometry if there are multiple ones?

cportele:
Yes, but I do not understand how this concept applies to a coverage.

The changes in #1003 simply replace "feature" in a few places with a more generic term, but all the other content is still about features. If the requirements class can be generalized, it needs to be changed to very clearly specify how these concepts relate to other data that is not feature data.

cportele:
Addressed by #1009

Issue #997, opened 2025-05-31
CQL2 grammar: BNF definition of function


dannbuckley:
Hello,

I am currently building an implementation of the CQL2-Text spec and had a question about the current definition of the `function` rule in the BNF schema.

https://github.com/opengeospatial/ogcapi-features/blob/64e26b624a052510b7d10e93b304cad0e5b30194/cql2/standard/schema/cql2.bnf#L355

Based on the rest of the schema, you seem to be using the `{ ... }` brace syntax to signify "1 or more". So, from my current understanding, wouldn't the current rule imply that something like `function(arga1, arga2, arga3 <space> argb1, argb2, ...)` is valid, with more than one argument list provided?

The closest similar rule I could find is `isInListPredicate`:

https://github.com/opengeospatial/ogcapi-features/blob/64e26b624a052510b7d10e93b304cad0e5b30194/cql2/standard/schema/cql2.bnf#L101

Here, the `inList` rule appears without braces and the resulting rule when substituted would be a single instance of `scalarExpression [ {"," scalarExpression} ]`

I'm wondering whether the `argumentList` in `function` should have its braces removed, giving something like:

```diff
- function = identifier "(" {argumentList} ")";
+ function = identifier "(" argumentList ")";

# When substituted, this gives:
function = identifier "(" argument [ { "," argument } ] ")";
```

Thanks in advance!

kunlinyu:
@dannbuckley  I think this is just a typo.
This is my Implementation:
https://github.com/IndoorSpatial/cql2cpp/blob/7331c03bf0d9cc0dbcc7c518ef096d7df8020434/src/cql2_parser.y#L353

PR #1007, opened 2025-06-09
CQL2: Change function argument list to optional group

dannbuckley:
See issue #997

### DIFF
--- a/cql2/standard/schema/cql2.bnf
+++ b/cql2/standard/schema/cql2.bnf
@@ -352,7 +352,7 @@ identifierStart = "\x003A"              # colon
 #=============================================================================#
 # Definition of a FUNCTION
 #=============================================================================#
-function = identifier "(" {argumentList} ")";
+function = identifier "(" [argumentList] ")";
 
 argumentList = argument [ { "," argument } ];
 


dannbuckley:
Hi @pvretano 

I agree that the function rule should allow for zero or more arguments.

In my experience, and from what I've gleaned from the current CQL2 semantics, `{braces}` usually mean one or more whereas `[square brackets]` indicate an optional group (zero or one occurrences). To express the "zero or more" case, current CQL2 rules use both types of brackets in conjunction. Here are a few examples:

Boolean expression, with zero or more boolean terms
```
booleanExpression = booleanTerm [ { "OR" booleanTerm } ];
```

List predicate, with zero or more additional list items
```
isInListPredicate = scalarExpression ["NOT"] "IN" "(" inList ")";
inList = scalarExpression [ { "," scalarExpression } ];
```

Character literal with zero or more characters
```
characterLiteral = "'" [ {character} ] "'";
```

If we interpret `{braces}` to mean one or more, that means (a) that current functions require at least one argument, and (b) that functions can have multiple argument lists delimited by whitespace. If the function rule is updated to instead use `[square brackets]`, then the first argument becomes optional and the function is limited to only having one argument list.

The expanded function rule would look something like this:
```
function = identifier "(" [ argument [ { "," argument } ] ] ")";
```

Issue #954, opened 2024-09-24
Part 5: OGC API - Common - Part 2 "Schemas" req.class Syncing


jerstlouis:
As previously discussed in the SWG, here is the feedback from the OGC API - Common - Part: Geospatial Data "Schema" requirement class integration.
Sorry for the late feedback, but I hope this can still be considered as a response to the RFC.

This Pull Request aligns with the initial changes we made for the OGC API - Common - Part 2: Geospatial Data Schemas requirement class:

https://github.com/opengeospatial/ogcapi-features/pull/953

The two main changes in this PR relate to being able to describe the meaning of enumeration values and nil values.
It also reference QUDT in general instead of units specifically, since QUDT can also be used for `x-ogc-definition`.
The remaining changes are mostly grammatical / editorial.

This PR only affects the first section 7 - Schemas of Features Part 5.

I've also just integrated the Core Roles, Returnables, Queryables and Sortables in this commit:

https://github.com/opengeospatial/ogcapi-common/commit/11d6bfdb86f4af1eb9adea4a7639f6fac1ad42a7

https://github.com/opengeospatial/ogcapi-common/commit/11d6bfdb86f4af1eb9adea4a7639f6fac1ad42a7#diff-3e75532a213bcba1f4d083a089ebc2dc5217a3f743e1950b6c4767f14d4195ec

currently organized into:

[clause_14_schemas.adoc](https://github.com/opengeospatial/ogcapi-common/blob/11d6bfdb86f4af1eb9adea4a7639f6fac1ad42a7/collections/clause_14_schemas.adoc)

and

[requirements_class_schemas.adoc](https://github.com/opengeospatial/ogcapi-common/blob/11d6bfdb86f4af1eb9adea4a7639f6fac1ad42a7/collections/requirements/requirements_class_schemas.adoc)

I will likely re-organize the files into individual requirement files to follow how things are organized elsewhere in Common.
The version in this specific commit might be easier to compare with.

The approach I took here was to merge requirements into separate parts of single requirements for brevity, and to merge Schemas, Core Roles, Returnables, Queryables and Sortables into a single "Schemas" requirement class for Common - Part 2.

At least for Common I think this is simpler, and there is not really a need to clients to declare conformance to these things separately. I believe the occurence (or not) of the associated specific properties or link relations really is enough to know that these things are supported by the implementation.

[...]

Issue #938, opened 2024-07-09
Part 5: Feature references with more complex values


cportele:
Part 5 (Schemas) currently specifies that feature references are a string or integer which is the local feature id of the referenced feature. See https://docs.ogc.org/DRAFTS/23-058r1.html#rc_feature-references.

This approach does not fully support cases where the feature id is not the only variable part of the URI of the target resource, for example, the collection id of the feature collection.

In those cases, the feature reference might have to be an object with one member for every parameter in the target URI template. For example, if there are multiple types (collections) of road segments, it could be necessary to use something like `{ "roadSegment": { "featureId": 123, "collectionId": "slip-road" } }` instead of `{ "roadSegment": 123 }` to be able to mint the URI of the target resource.

This would also impact queryables for such properties and how to write filter expressions.

It could also be an option to consider such feature references as too complex and out-of-scope for Part 5. In that case implementations / communities that need such a capability would have to specify their own extension.

cportele:
Meeting 2024-08-26: Cubewerx includes the collection information in the featureId/key, e.g., https://www.pvretano.com/cubewerx/cubeserv/default/ogcapi/zaatari/collections/SettlementSrf/items/CWFID.SETTLEMENTSRF.0.6. In ldproxy two properties (one for the featureId and one for the collection). We need to analyze/discuss, if we need to specify something in Part 5.

cportele:
We decided to consider more complex use cases as out-of-scope for v1.0

PR #1009, opened 2025-07-14
Prepare Part 5 for v1.0.0

cportele:
I have reviewed the current draft with the goal to

- make the req class for property roles for features less feature-specific;
- resolve/remove open issues;
- improve consistency / find errors.

The pull requests #1002 and #1003 and the issue #954 should be closed when this pull request is merged.

The open issue #938 can also be closed (as discussed in the last meeting, I have removed more complex references from the scope).

If the pull request is merged, there are two additional changes that are needed before submitting the document to the TC:

- a third submitter (with TC voter status);
- the Abstract Test Suite.

Finally, a topic that came up in our implementation that I also wanted to raise before we finalize the document: Currently we have only brief discussion in 5.3 on schemas for validation (to use `rel=describedby` to link to them). Is this sufficient or do we want to recommend a path for such schemas?

### DIFF
--- a/extensions/schemas/standard/clause_0_front_material.adoc
+++ b/extensions/schemas/standard/clause_0_front_material.adoc
@@ -73,7 +73,10 @@ Recipients of this document are requested to submit, with their comments, notifi
 The following organizations submitted this document to the Open Geospatial Consortium (OGC):
 
 * CubeWerx Inc.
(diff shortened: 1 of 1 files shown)


cportele:
Meeting 2025-07-14: Add example for a reference with the URI template and then merge.

PR #1035, opened 2026-04-13
Initial checkin.

pvretano:


cportele (approved):
I made some editorial changes and added an entry so that the HTML draft is generated nightly.

I noticed some issues (e.g., the JSON Schema is specific to features, but is intended to be generic, I think), but we should discuss them in the process of upgrading the proposal to a draft.

Issue #1070, opened 2026-08-20
Part 4: Outcome of asynchronously processed mutation requests (status code 202)


cportele:
The four response requirements state that a server returns 202 if an operation is added to a processing queue, and the accompanying prose notes that "the processing can succeed or fail, without further notification to the client."

There is currently no way for a client to determine the outcome. In particular, for a CREATE request the Location header is only required for 201 responses, so after a 202 the client does not even know the URI of the prospective resource and cannot poll for it. There is no way for an Abstract Test Suite to test 202 responses.

Options: 

1. require or recommend that a 202 response to a CREATE request include a Location header with the URI of the prospective resource, so clients can poll; 
2. leave Part 4 as-is and defer richer async handling to Part 11 for atomic/batch transactions, which includes an "Async transactions" requirements class;
3. both — (1) as a minimal improvement, with a note pointing to (2); or
4. do not allow 202 as a response to simple Create, Replace, Update or Delete requests acting on a single resource.

cportele:
Meeting 2026-08-24: Allow 202 (permissions), but do not go into detail about the jobs and async request handling. If a Jobs standards becomes available in time as a separate standard (co-branded with Common), we should reference Jobs and keep the 202 and add the `respond-async` preference. This could also be in a follow-up version 1.1.

PR #1072, opened 2026-08-23
Part 4: Add Abstract Test Suite plus executable tests

### LINKED ISSUE #1071: Part 4: Create Abstract Test Suite
Annex A currently states:

> Conformance classes and test cases for all requirements classes and requirements will be added once the requirements classes and requirements are final.

cportele:
Closes #1071.

Depends on #1069. This PR is stacked on the review PR and should be merged after it.

New issue: #1070.

## Summary

This adds the Abstract Test Suite as Annex A, a test dataset, and executable tests
that run the suite against a Web API. Writing and executing the tests exposed a
number of issues in the requirements — untestable conditions, a missing requirement,
two requirements that contradicted each other, and three provisions that contradicted
HTTP. Those are fixed here, because the tests cannot be written consistently
otherwise. All changes outside Annex A are listed below with their reason.

## Annex A: Abstract Test Suite

One conformance class per requirements class, 34 tests in Given-When-Then notation
(the style of OGC API - Features - Part 5: Schemas and of CQL2):

| Conformance class | Tests |
| --- | --- |
| Create/Replace/Delete | 8 |
| Update | 4 |
| Optimistic Locking using Timestamps | 4 |
| Optimistic Locking using ETags | 4 |
| Handling Preference | 4 |
| Features | 10 |

Conventions that the introduction of the annex establishes, because the tests of this
Standard change resources and therefore cannot be written like tests of a read-only
API:

* a **"Mutable Resources" test parameter** — which resources are mutable, which
  operations each of them supports, and a valid request body per operation; this
  depends on the API and on the credentials used by the test, so it cannot be
  discovered;
* a **"Condition" row in every test**, evaluated against the conformance declaration
  and the test parameters, so that a test that does not apply is skipped rather than
  failed. The conformance class "Features" mirrors the conditional dependencies of
  its requirements class;
* how the **media types supported in a request body** are determined (the
  `Accept-Post` and `Accept-Patch` headers of the OPTIONS response, a `415` response
  as the fallback);
* that a server implementing an Optimistic Locking class **may require a
  precondition** in every state-changing request, so the tests of the other classes
  state one, obtained with a preceding GET — and state none for a resource that does
  not exist yet;
* that assertions about the state of a resource are **skipped for a `202` response**,
  since this Standard specifies no mechanism to determine the outcome of a queued
  operation.

## Test dataset (`standard/data`)

Eleven files: three building features, the logical schema of the collection, request
bodies for CREATE (valid, invalid, JSON-FG), REPLACE and UPDATE, the same in GML, and
the GML application schema. The `updated` property carries the timestamp of the last
change, from which a server derives the `Last-Modified` header, so that the
Optimistic Locking tests can be executed against the dataset.

## Executable tests (`executable-tests`)

A test runner that implements all 34 tests (Python 3 standard library only) and a
deployment that publishes the test dataset with ldproxy on PostGIS:

```sh
cd extensions/transactions/create-replace-update-delete/executable-tests
./run-tests.sh                  # all applicable tests
./run-tests.sh --verbose        # log every request and what it tests
```

The deployment provides two APIs and three collections, so that the conditions of all
tests are met by at least one of them: a collection with server-assigned feature
identifiers and optimistic locking, a collection with client-assigned identifiers
(features created with PUT), and an API without support for coordinate reference
systems other than WGS 84 longitude/latitude — the conformance declaration is per API,
so the condition of `/conf/features/crs84` cannot be met by a second collection.

The suite was executed against ldproxy while it was being updated in parallel: 30
of the 34 tests pass, the four that are skipped are the tests of "Optimistic
Locking using ETags", the only conformance class the implementation does not support.
Every other requirement of this Standard has therefore been verified against an
implementation at least once.

These are not an official Executable Test Suite of the OGC; they exist to verify that
the tests are executable and the requirements testable.

### New requirements

| Requirement | Reason |
| --- | --- |
| `/req/update/patch-content-type` | The Update class had no rule for the `Content-Type` header of a PATCH request, although `/per/update/content-type-lenient` presupposed one. Mirrors `/req/create-replace-delete/post-content-type` and `.../put-content-type`. |
| `/req/optimistic-locking-timestamps/ifunmodifiedsince-delete-op`, `.../ifunmodifiedsince-delete-eval`, `/req/optimistic-locking-etags/ifmatch-delete-op`, `.../ifmatch-delete-eval` | Both Optimistic Locking classes specified conditional processing only for PUT and PATCH. Deleting a resource that another client has changed in the meantime is the same lost-update problem, and RFC 9110 preconditions apply to every state-changing method. |
| `/req/create-replace-delete/options-accept-post`, `/req/update/options-accept-patch` | Three conditions in the Features class required the server to "advertise the media type in the API definition", which is not easily testable unless a particular API definition is assumed. The media types are now discoverable from the API itself, with `Accept-Post` (W3C Linked Data Platform 1.0) and `Accept-Patch` (RFC 5789, section 3.1). |
| `/req/features/schema` provision D | The requirement stated only that a server must not reject unspecified properties *unless* the schema says `"additionalProperties": false`. It did not say what happens if the schema does say so, so a server could publish that constraint and silently drop such properties, which a client cannot detect. The example in clause 10.6 already assumed the rejection. |
| `/req/features/schema` provision E | Part 4 used the term "receivable property" without saying how a client determines the receivable properties, and nothing covered a request that *sets* a read-only property: provision D covers only properties that the schema does not specify at all, and `/req/features/update-json-merge-patch` B only the property with the role `id`. A server-maintained property such as a timestamp of the last change is in the same situation — in the test dataset, `updated`. Provision E requires the rejection, and clause 10.6 explains where the receivable properties come from. |

### Changed requirements and permissions

| Provision | Change and reason |
| --- | --- |
| `/req/create-replace-delete/put-response` | Distinguishes replacing the content of an existing resource (`200`/`204`) from creating a resource (`201`). Section 9.3.4 of RFC 9110 requires `201` in the latter case: "If the target resource does not have a current representation and the PUT request successfully creates one, the origin server MUST send a 201 (Created) response". The requirement previously allowed only `200` or `204` for any successful PUT, which made the upsert case non-conformant with HTTP. |
| `/per/create-replace-delete/put-create` | States that the response of such a request is `201` and that no `Location` header is needed, because the resource is identified by the request URI (RFC 9110, section 15.3.2) — POST-create does require `Location`, so the asymmetry would otherwise look like an omission. |
| `/req/create-replace-delete/put-rid-exception` B | Now applies only to a server that supports creating resources with PUT. For a server that does not, the request is unsuccessful in any case and provision A already requires `404`; section 13.2.1 of RFC 9110 gives such a failure precedence over the evaluation of a precondition. A note explains the interaction. |
| `/req/create-replace-delete/put-rid`, `/req/update/rid` | Both required the server to *ignore* a resource identifier in the request body, while `/req/features/update-json-merge-patch` B requires a request that updates the property with the role `id` to be *rejected* — a server could not satisfy both. Both now permit either behaviour and leave the choice to the specification of the resource type, which is what the tests already assumed. A note in clause 7.2 explains why the two operations differ: in a REPLACE operation the identifier in the body is redundant, while in an UPDATE operation the body describes the changes, so a different identifier requests a change of the identifier. |
| `/req/features/geojson-create-replace`, `/req/features/gml-create-replace`, `/req/features/update-json-merge-patch` (conditions) | The conditions referenced the API definition (see above) and now refer to the media types that the server supports in the request body. The condition of the GML requirement also said `application/gml`, which is not the GML media type. |
| `requirements_class_create-replace-delete` | Added the dependency on W3C Linked Data Platform 1.0 for the `Accept-Post` header field. |

### Removed

A number of outdated and orphaned files were deleted.

### OpenAPI examples

`examples/openapi` now contains the OpenAPI definitions of the two APIs of the test
deployment, exported from the running deployment, as examples of how the resources and
operations of this Standard can be described in an API definition — including both ways
of creating a feature (POST for server-assigned identifiers, PUT for client-assigned
ones) and the `Content-Crs`, `Prefer` and profile-`Link` parameters. The README notes
that they are not normative.


pvretano (approved):
LGTM ...

PR #1080, opened 2026-09-05
Part 4: clarify 202 responses

### LINKED ISSUE #1070: Part 4: Outcome of asynchronously processed mutation requests (status code 202)

cportele:
Make 202 (asynchronous processing) a permission, add note on future respond-async/jobs support.

Closes #1070.


pvretano (approved):
LGTM

PR #1081, opened 2026-09-05
Part 4: add pointers to introductory/reference material on Web authentication

### LINKED ISSUE #694: Add links to external resources that provide an introduction to Web authentication
Meeting 2022-05-23: This topic is not specific to geospatial, so the current thinking is that we should not add much in the OGC API standards, but we should provide links to good introductory material on the subject that introduce JWT, OAuth2 flows, OpenIDConnect, etc to geospatial developers. Maybe @securedimensions can provide some input?

@pvretano will check, if Cubewerx can provide introductory text from their implementation experience.

The July Code Sprint will include OGC API Features and we should consider to make Part 4 with security a focus of the Code Sprint for Features.

cportele:
Closes #694

@pvretano - I compiled a list of introductory or reference material, just in case you did not find the time to work on this. Otherwise just close this PR.


pvretano (approved):
LGTM.

PR #1082, opened 2026-09-13
Part 1: move the description of links into a dedicated sub-clause

### LINKED ISSUE #918: A dedicated section for links
I was just looking for a dedicated section that describes the concept (and schema) of a link in OGC API - Feature - Part 1.
More specifically, I was looking for the link schema in GeoJSON and the first clickable link to the schema is in 7.10. String internationalization, which is rather weird.
The term link is generally used various times throughout the document, but the actual information about links (schemas, relation types, ...) are scattered around. 

Examples:
- [5.2. Link relations](https://docs.ogc.org/is/17-069r4/17-069r4.html#_link_relations)
- [7.12. Link headers](https://docs.ogc.org/is/17-069r4/17-069r4.html#_link_headers)
- [7.10. String internationalization](https://docs.ogc.org/is/17-069r4/17-069r4.html#string_i18n)
- 8. Requirements classes for encodings

I think it would be gread if there would be a central section that describes the specifics of the links:
- RFC 8288 dependency
- Schemas for the encodings
- Use of relation types
- etc.

It might just be a summary somewhere at the beginning with links to the corresponding sections, but I think a central place that people can link/point to would be good.

Edit: Common has something similar: https://docs.ogc.org/is/19-072/19-072.html#_11b9b4f7-42fc-413a-b63a-e7fb060b5e4b

cportele:
Closes #918

Web linking was scattered over clause 5.2 (link relations), 7.12 (string internationalization) and 7.13 (link headers), and the link schema was only reachable via a $ref. Collect it in a new sub-clause 7.2 "Links", placed before the landing page since that clause already discusses links:

- 7.2.1 Link objects: the RFC 8288 dependency, the link schema inline, link templates, and how links are represented in each encoding
- 7.2.2 Link relations: the list moved from clause 5.2
- 7.2.3 Link headers: unchanged, /rec/core/link-header
- 7.2.4 Links to resources in other languages: the hreflang paragraph moved from string internationalization

The sub-section titles keep the auto-generated anchors _link_relations and _link_headers, so existing deep links continue to resolve.

Also corrects a broken cross-reference (rfc8228 -> rfc8288), which was in the moved hreflang paragraph.


cportele:

PR #1083, opened 2026-09-13
Part 1: add the "profile" link attribute to the Link object

### LINKED ISSUE #1031: Add "profile" link attribute to the Link object
From https://github.com/opengeospatial/ogc-feat-geo-json/issues/129:

> We will add "profile" as an additional link attribute as a hint indicating what the profile of the result of dereferencing the link should be. It should only be used in combination with the "type" attribute (since the value identifies the profile of a media type). The "profile" link attribute should be added to the OGC Link schema consistently. In Features Part 1 we will add it in the planned revision 1.1.

This also aligns Features with OGC API Records. 

In Features this should be used in at least the following links:

- `self`/`alternate` links
- `items` links from the collection

The "profile" link attribute is especially important in cases where different profiles of a format / media type are important to distinguish. The main cases in OGC standards at the moment are the GeoJSON profiles (`rfc7946`, `jsonfg`, and `jsonfg-plus`).

In general multiple profiles may apply (like in the `profile` query parameter, which is an array). 

In a `Link` header, the profile URIs have to be concatenated, separated by whitespace (like in the link attribute "rel").

In a JSON Link object "profile" is an array, too. See [Records](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/linkBase.yaml).

Example in a collection:

```json
{
  "href": "/collections/foo/items?f=json&profile=rfc7946",
  "type": "application/geo+json",
  "profile": [ "http://www.opengis.net/def/profile/ogc/0/rfc7946" ],
  "rel": "items",
  "title": "The 'foo' features in GeoJSON"
},
{
  "href": "/collections/foo/items?f=json&profile=jsonfg",
  "type": "application/geo+json",
  "profile": [ "http://www.opengis.net/def/profile/ogc/0/jsonfg" ],
  "rel": "items",
  "title": "The 'foo' features in JSON-FG"
}
```

cportele:
Closes #1031

Adds "profile" to the link schema as a hint about the profiles that apply to the target resource, aligned with the OGC Link schema in OGC API - Records. The attribute is only meaningful together with "type", since a profile is always a profile of a media type, and it is a list because more than one profile can apply.

- linkBase.yaml, so both link.yaml and linkTemplate.yaml inherit it, and the resolved link/linkTemplate components in the OpenAPI 3.0 and 3.1 bundles
- the link schema and the discussion in 7.2.1, with an example of the GeoJSON profiles in the "items" links of a collection
- /rec/core/link-profile, including the whitespace-separated form of the "profile" link parameter in a Link header
- the term "profile" in clause 4 and RFC 6906 in the bibliography, using the same definition as Part 5

No requirement changes: the requirements reference the link schema by $ref, so the optional attribute flows through.

Note that the profile URIs in the issue use a lower-case authority. Only the upper-case form is registered with the OGC NA, so the document uses http://www.opengis.net/def/profile/OGC/0/rfc7946.


cportele:
Meeting 2026-09-21: Maybe we need to sync with other OGC API standards that have extended the base link object, e.g.  https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/linkBaseExtended.yaml

PR #1086, opened 2026-09-13
Part 1: add the page size of the features to the collection

### LINKED ISSUE #538: Consider adding a new collection property to recommend a paging size
Here are some typical values for the “minimum”, “maximum”, and “default” properties of the [“limit parameter"](http://docs.opengeospatial.org/DRAFTS/17-069r4.html#_parameter_limit):

- minimum: 1
- maximum: 10000 
- default: 10

None of these properties can help client implementations that rely on picking a good “limit” for the initial load/display of a collection. These properties affect all collections in an OAFeat service. The default limit of 10 features is too low. Can 3000, 5000, or the maximum of 10,000 be the best paging size that fits all collections? While a paging size of 10,000 works well for a 50,000 point collection, it might not be ideal for a 50,000 polygon collection. 

Adding a new property e.g. “RecommendedPagingSize” at the collection level would provide a hint to clients on what could be an ideal paging size for each collection. This removes the guessing factor and makes it the responsibility of the service creator to define this property.

cportele:
Closes #538 

Adds recommended "defaultItemsPageSize" and "maximumItemsPageSize" properties to the collection. Both are also in the API definition, but extracting them from it is error-prone and they can differ per collection, so a client currently has to send a request with a large "limit" and derive the maximum from the "next" link.

No "recommended page size" is added: what a good page size is depends on the client and the network, not only on the collection.

- the properties in collection.yaml and in the collection component of the OpenAPI 3.0 and 3.1 bundles
- /rec/core/fc-md-items-page-size, including that the values have to be consistent with the "limit" parameter in the API definition


cportele:

PR #1087, opened 2026-09-13
Part 1: reword /rec/core/fc-md-item-link-templates

cportele:
Part B stated a requirement with SHALL inside a recommendation. Merge it into part A, so that the recommendation only recommends, and renumber the remaining part. The statement is unpublished yet.


cportele:

PR #1088, opened 2026-09-13
Part 1: add a discussion of CURIEs

### LINKED ISSUE #632: Support for CURIEs (compact URIs)
The OGC Naming Authority assigns URIs to resources that are registered (like coordinate reference systems, link relation types, etc.). During the September 2021 OGC Member Meeting the OGC Naming Authority has decided to allow the use of [CURIEs](https://www.w3.org/TR/curie/) to allow for shorter URIs with the following conventions (see [the comments starting with this comment](https://github.com/opengeospatial/NamingAuthority/issues/92#issuecomment-888186821)):

A CURIE `{authority}[-{objectType}]:{id}` would map to the following OGC URI: 

`http://www.opengis.net/def/{objectType}/{authority}/0/{id}`

If `-{objectType}` is missing, the default object type is `crs`.

That is, we could use, for example,

* `ogc-rel:conformance` as an alias for `http://www.opengis.net/def/rel/ogc/0/conformance` 
* `epsg:4326` as an alias for `http://www.opengis.net/def/crs/EPSG/0/4326`
* `ogc:CRS84h` as an alias for `http://www.opengis.net/def/crs/OGC/0/CRS84h`

We should consider to allow the use of these CURIEs for CRS URIs (at least in the query parameters "crs", "bbox-crs" and "filter-crs", but maybe also in the "crs" and "storageCrs" members in the Collection resource) and OGC link relation type URIs ("rel" member in Link objects, with a default for "ogc-rel" as the prefix). At least for "rel" the CURIEs should be safe CURIEs (in square brackets) to avoid ambiguous cases, because "rel" values are often not URIs.

This would affect parts 1, 2 and 3 and will require an update to the schema for the relevant building blocks. In case of parts 1 and 2, this will require a version 1.1. It might also be an option to just add normative statements  in an additional requirements class that allow the CURIE values in specific building blocks.

As an aside, there is an open issue that not all resources have a OGC URI with version "0" (e.g. CRS84). The OGC Naming Authority is still looking into [how to address this in a consistent way](https://github.com/opengeospatial/NamingAuthority/issues/120).

cportele:
Closes #632 

Adds a sub-clause "CURIEs" to the conventions: what a CURIE is, that OGC API Features does not use CURIEs in resource representations, and that whether a CURIE can be used as the value of a query parameter is a question for the parts that specify a parameter whose value is a URI.

Keeping URIs in resource representations means that a client can compare a value against the URIs it knows without expanding a CURIE first. A query parameter value is processed by the server only, so supporting CURIEs there does not affect clients that do not use them.


cportele:

PR #1091, opened 2026-09-13
Part 1: clarify that the collection extent includes its boundary

### LINKED ISSUE #934: Collection extent: End inclusive or exclusive
Is the end timestamp that is specified in the extents.temporal.interval property of a Collection as second element of each subarray inclusive or exclusive? I can't find it defined in the specification.

cportele:
Closes #934 

It was not stated whether the ends of a bounding box or of a time interval in the extent of a collection are inclusive or exclusive. They are inclusive: an extent that is derived from the features in the collection normally has features on its boundary.

/req/core/fc-md-extent-multi stated that each feature is "inside the extent described by" a bounding box or a time interval. The four parts now state that every coordinate value is in the closed range that the bounding box defines for the axis, and that every instant is in the closed time interval.

The antimeridian case and half-bounded intervals are covered in notes. The convention that the west-most value of a bounding box is the larger one where the box spans the antimeridian is already stated in this clause, and that a null value indicates a half-bounded interval is already stated in /req/core/fc-md-extent.

The requirement keeps its identifier and the number of parts, so the Abstract Test Suite is unchanged: /conf/core/fc-md-extent-multi refers to the requirement as a whole.


cportele:

